PR #13, opened 2025-12-24
string splitting

TheLDB:
idk abt function name `str_split` maybe could be better


src/vm/vm.v
```diff
@@ -774,6 +775,25 @@ fn (mut vm VM) execute() !bytecode.Value {
...
+				if string_val is string && delimiter is string {
+					parts := string_val.split(delimiter)
+					mut result := []bytecode.Value{cap: parts.len}
+					for part in parts {
+						result << part
+					}
+
+					vm.stack << bytecode.Value(result)
+				} else {
+					return error('str_split requires string and delimiter')
```
alii, line 794: This is fine for _now_ but I do want to never have any cases the vm can panic. This means that the signatures of stdlib functions need to represent the failure cases as well. 

With that said, this failure case would be caught by the checker. However that's not a silver bullet because these two code paths have to stay "in sync" to agree. If AL ever goes 1.0 I would want to find a more intelligent way to define builtins and eliminate classes of bugs entirely where the vm could panic

alii (approved):
THANK YOU

PR #16, opened 2025-12-26
multiline comments + lsp qol 

TheLDB:
- added multiline comment (`/* */` and `/** */`) parsing
- updated al.tmLanguage.json to support multiline comments
- updated parser to extract doc comment trivia and attach it to the node
- added docs mapping `map[string]string` to the environment (so usages / call sites can look up definitions 
- updated lsp send_response to take in a generic and actually encode a struct to send back (prev was just handwriting strings)
- added hover support to see doc comments (`/** */`)
- added go to definition for struct field uses / enum variant uses
- deprecated commas on struct fields

test file at `program/src/doc_comment.al`


src/scanner/scanner.v
```diff
@@ -78,6 +78,34 @@ fn (mut s Scanner) collect_trivia() {
...
 
+		// Collect block comments /* */ and doc comments /** */
+		if ch == `/` && s.state.get_pos() + 1 < s.input.len && s.input[s.state.get_pos() + 1] == `*` {
+			start := s.state.get_pos()
+			s.incr_pos() // skip /
+			s.incr_pos() // skip *
+
+			// Doc comment if next char is * but NOT followed by /
+			// i.e., /** but not /**/ or /***/
+			is_doc := s.state.get_pos() < s.input.len && s.peek_char() == `*`
```
alii, line 89: Nice!

alii (approved):

PR #23, opened 2026-08-10
Tree-sitter grammar for Scarlet, single-sourced editor lexicons

alii:
## What this adds

- `cargo xtask gen-editor-syntax` generates `scarlet.tmLanguage.json` and `editors/tree-sitter-scarlet/lexical.js` from the compiler's token tables: `Keyword::ALL`, the fixed spellings of `Kind`, and the scanner's escape table. A new token kind does not compile until it is classified for the editors. `cargo test -p xtask` fails while the generated files are stale.
- `editors/tree-sitter-scarlet/` is a tree-sitter grammar transcribed from the reference parser: the full precedence ladder, every pattern form, bit-string literals and patterns with `:size/:bytes/:binary/:utf8` specs, labeled arguments, spreads, backpassing, `or` with binder, and nested string interpolation. `src/scanner.c` produces one token: a `-` at the start of a line, the parser's fresh-line-minus rule. The LR grammar cannot express it.
- `bun run corpus` parses every `.scrl` file in the repository — 72 files across examples, stdlib, and test programs — and fails on any error node. CI runs it, and fails when `src/` is stale relative to `grammar.js`. The grammar tracks the compiler or the build fails.
- The Zed extension declares the grammar and ships highlight, bracket, indent, and outline queries.

## Corrections

The previous tmLanguage highlighted `assert`, `true`, `false`, and backtick character literals. The language has none of them. The generated file adds `opaque`, bit-strings, and the exact escape set. It marks `++`, `--`, `;`, and `?` as invalid: the scanner produces those tokens, and no parser rule consumes them.

## Verification

- `cargo fmt` and `cargo clippy --all-targets` are clean. `scarlet_syntax` and `xtask` tests pass. The 8 `core_ir` golden failures predate this branch.
- The release CLI — `check`, `run`, and the REPL — produces byte-identical output for every escape and for nested interpolation. Unknown-escape diagnostics are unchanged.
- Generated artifacts (`src/`, `lexical.js`, the tmLanguage) are marked `linguist-generated`. The hand-written surface is about 1,700 lines.


pxseu:
Thanks this will be really cool and will unblock me. 🩷

PR #27, opened 2026-08-12
scarlet/json: simd parse to a tape, typed decoders

alii:
`scarlet/json` gives Scarlet programs a JSON parser and a typed decoder. The stdlib had neither: grepping `crates/scarlet_core/src/std/` for "json" returned nothing.

## The crate

`simd-json` 0.17.3, `runtime-detection`, `default-features = false`. One binary serves SSE4.2, AVX2 and NEON, so the org's arm64 machines are covered. `to_tape()` returns a flat node array with subtree counts, which is the substrate this design needs.

`sonic-rs` was not benchmarked. That comparison is open.

## The seam: a tape, not a tree

`json.parse` returns an opaque `Doc` — a cursor over a compact 16-byte-per-node tape plus a string arena, both handed across as `Binary`. Nothing materialises until it is read. That is the answer to "the boundary is where the throughput goes".

**Nothing recurses on the parse path.** simd-json accepts 10,000-deep nesting, measured, so a recursive tape build or encoder would be a reachable stack overflow. Tape build is one linear pass; the encoder carries its own work stack.

## Absent, null and present are three states

`scarlet/json/field.Field(a)` is `pub opaque`. Constructors are unexported, so it cannot be matched on from outside and cannot be collapsed by accident. There is deliberately **no `Field(a) -> Option(a)`**. Two ways through:

- `fold(f, on_absent, on_null, on_present)` — one argument per state
- `update(f, current Option(a))` — the partial-update rule, which takes the value it must preserve, so the "unchanged" branch has something to leave

`decode.optional_field` returns `Field(a)`, so the familiar name leads to the safe type.

Stronger than intended: **planting the collapse (`Absent -> None`) does not compile.** It leaves `current` unused and Scarlet rejects that.

## Verification

`cargo fmt --all` 0. `cargo clippy --all-targets` 0, zero warnings. `cargo test` 0 — **41 suites, 1224 passed, 0 failed**.

Five plants, each watched failing and restored:

1. surrogate pair consumed as one escape → `lone_surrogates_are_found_and_valid_pairs_are_not`
2. container skip using `count` not `count+1` → `skip_walks_past_a_nested_object`
3. float via `Display`, so `1.0` prints `1` → `float_encoding_round_trips_and_keeps_its_kind`
4. `u64 > i64::MAX` truncated → `integer_above_i64_max_is_kept_distinct_not_truncated`
5. `Null -> current`, so a null stops clearing → golden: `want "after null None", got "after null Some(lovelace)"`

## Numbers — x86_64, release, min of 3

Payloads byte-identical on both sides. **loadavg 1.05 on 16 cores at start: a fleet was live, so these are lower bounds.**

| | small, 141 B | large, 413,707 B |
|---|---|---|
| Scarlet parse → `Doc` | 0.75 µs | 1.4 ms — **295 MB/s** |
| Scarlet parse + decode | 5.6 µs | 25.0 ms |
| `serde_json` → `Value` | 0.58 µs | 3.69 ms — 112 MB/s |
| `serde_json` → typed struct | 0.25 µs | 0.855 ms — 484 MB/s |
| `simd-json` → `OwnedValue` | — | 3.2 ms — 129 MB/s |

Three findings:

1. **On a large document the tape beats both tree builders** — 2.6x `serde_json`→`Value`, 2.3x `simd-json`→`OwnedValue`.
2. **On a small document SIMD does not win.** Scarlet's parse is 1.3x *slower* than `serde_json`→`Value` at 141 bytes; fixed cost dominates. This is a real finding, not a failure.
3. **The decoder is 94% of end-to-end**, 25.0 ms against 1.4 ms. An O(n²) bug in the benchmark harness was found and fixed first; it moved the number 2%, so the finding stands.

## One correctness call worth reviewing

**`parse` rejects unpaired UTF-16 surrogate escapes.** simd-json silently decodes `\ud800` to NUL, which is indistinguishable from a real NUL, so a caller comparing a decoded string against a token can be fooled. A pre-parse scan, memchr-gated, stops immediately on documents with no backslash. This diverges from the library and costs a pass. It is one guard to remove if the library's behaviour is preferred.

## Known gaps

- The READY payload in the tests is READY-*shaped*, hand-built from the public structure. It is not a real capture.
- `json.to_json` and `decode.list` recurse in Scarlet, so deep documents can exhaust the Scarlet stack on the materialising path. `parse` itself is safe. Untested past 600 deep at the Scarlet level.
- Core IR goldens moved — type ids shifted by the new stdlib types. Regenerated with `UPDATE_CORE_GOLDEN=1`; mechanical, 8 files.


alii:
Adversarial review returned **do not merge** on three blockers plus a finding that meant this PR was never actually reviewable. All fixed at `6185d44`.

**B4 — a raw NUL byte on line 178 made git classify the parser as binary.** GitHub showed `Binary file not shown` for 853 lines — the entire security-relevant surface — with no line-level review possible. `git grep` returned no line number; `rg` skipped it. `fmt --check` and `clippy` both passed. Fixed in its own commit (`990b034`) so the change that makes the file reviewable is itself reviewable. The diff now renders 1,112 lines of text.

**B1 — `json.elements` was quadratic**, an algorithmic DoS on any attacker-sized array, and the documented way to decode untrusted JSON. Measured, release, x86_64:

| | cost per 2× input | exponent over 8× |
|---|---|---|
| before | 3.92×, 3.98×, 3.99× | **n^1.99** |
| after | ~1.7–2.5× | **n^0.98** |

`collect_elements` is deleted; `elements` is now a single linear intrinsic mirroring `json_entries`. The in-program control (a pure Scarlet append loop) was flat before and tracks linearly after, confirming array append was never the cost.

**B2 — `{"id":18446744073709551615}` re-encoded to `{"id":0}`.** Fixed with `Json.Number(digits)` plus an intrinsic so the digits pass through verbatim — not as a truncation, and not as an approximate `Real`. `Number` is a public constructor, so the encoder validates at the boundary: text that is not a JSON number becomes `null` rather than corrupting the document around it.

**B3 — a decoder built only from optional fields accepted any document.** `[1,2,3]`, `"not a config at all"`, `7` and `null` all decoded `Ok` with every field `Absent`. Both `field` and `optional_field` now route through one `member` helper, so they cannot drift apart again. `field`'s output is byte-identical to before.

**B5 — the round-trip test compared the encoder to itself.** The shipped assertion printed `True` for the B2 input while `reencode` produced `{"id":0}`. The rewrite puts the input on one side, and printed `False` for that case **before** B2 was fixed.

**B6 — the fast-path test was decorative.** Now returns `(answer, steps)`. Planting the byte-at-a-time scan fails it `(None, 4096)` vs `(None, 1)` — **and the other 14 tests in the module still pass**, which was the review's point.

**B7 — `int_string` rejected negatives.** `"-5"` now decodes. `i64::MIN` still fails and is documented with why, rather than hand-rolling arithmetic that would replace `parse_int`'s established overflow detection.

A dead-code warning caught a dispatch site that would otherwise have been missed: the JIT bridge in `native_shims.rs` has a second `opc::` match beyond `exec.rs`. Both new ops are wired there.

Gates: `fmt --check` 0, `clippy --all-targets` 0 warnings, `test --workspace` 0 (41 binaries), `gen-editor-syntax --check` 0. The golden diff is purely additive — 31 new lines, no existing line changed.

**One correction to this PR's own claim.** "Collapsing the three states does not compile" overreaches. `fold` is exported, so `fold(f, fn() None, fn() None, fn(x) Some(x))` is exactly `Field(a) -> Option(a)` and compiles — and `decode.with_default` ships such a collapse. `field.scrl` line 18 says *"there is deliberately no `Field(a) -> Option(a)`. That function is the bug."* Line 46 is the honest one: *"a collapse is something you write on purpose and a reviewer can see."* **The guard is visibility, not type-level impossibility.**

Filed rather than folded in: `parse_with_limit` (100 MB parses to 2.26 GB maxrss, ~22×), the encoder's `[1,,3]` failure mode, `-1` reading as `ObjectKind`, the `syn` 3.0.3 duplicate, **no gate catches a NUL in source**, no gate holds `elements` linear, and `Doc`'s debug dumping the arena.

PR #28, opened 2026-08-12
int: bitwise and, or, xor, not and the two shifts

alii:
Scarlet had no integer bitwise operations. `BinaryOp` carried Add Sub Mul Div Mod Eq Ne Lt Le Gt Ge And Or, and `And`/`Or` are boolean short-circuit routed to `ShortCircuitOp`. `scarlet/int` exported five functions. A grep for `bitwise|BitAnd|BitOr|Xor|Shl|Shr` across the crates found nothing outside one Cranelift-internal `bxor_imm`.

This adds six functions on `scarlet/int`: `bitwise_and`, `bitwise_or`, `bitwise_xor`, `bitwise_not`, `bitwise_shift_left`, `bitwise_shift_right`.

## Functions, not operators

`or` is a keyword, so a bare `int.or` cannot parse. Once one name needs a prefix, a uniform prefix beats a special case.

Separately: a new infix operator is a permanent precedence commitment. C binding `&` looser than `==` is the canonical example of getting that wrong and never being able to fix it. Functions have no precedence to get wrong.

Names follow Gleam, since the port that surfaced this reads Gleam — except `bitwise_xor` rather than `bitwise_exclusive_or`, which is inherited Erlang baggage.

## Int is exactly i64

Verified rather than assumed: `bytecode/value.rs` uses 48-bit small integers spilling to a single-`i64` `BigInt`, and `ValueView::Int(i64)` collapses both. **Scarlet's ceiling is 2^63-1.** Callers porting from a language that caps decoders at 2^53-1 gain range rather than losing it.

## Total, and where that departs from C

Following the existing local precedent that `x / 0 == 0`, `x % 0 == x` and `abs` saturates: no traps, and no `Result` to thread through a bitfield expression.

- **A count of 64 or more shifts every bit out. The count is not masked to 6 bits.** `shift_left(1, 64)` is 0, not 1. This is a deliberate departure from C and from the x86 and ARM shift instructions, where the hardware masks and the answer surprises.
- **A negative count shifts the other way** by the same magnitude. `i64::MIN` goes through `saturating_neg`, which lands past the width — the same answer the exact magnitude would give.
- **`shift_right` is arithmetic.** Int is signed and has no unsigned twin, so `-8 >> 1 == -4`, and past the width a negative leaves `-1` rather than 0.
- **`not` complements the sign bit**, so `not(0) == -1` and `not(n) == 0-n-1`. This is the most surprising of the six and the docs say so, with the masking workaround.

## Seam

Six `@vm` builtins, six opcodes **appended** to the end of `Op` so no discriminant shifts, classified `OpCoverage::Bridge` — pure, single-result, cannot fail, the class `Add` and `Div` already sit in.

## Verification

`fmt` 0, `clippy` 0 with no warnings, `test` 0 — 41 test binaries, no failures.

Four plants, each watched red and restored:

1. **Masked the shift count** to C behaviour → `shift_past_the_width_does_not_mask_the_count` red (`left: 1, right: 0`), golden red: `want "shl 64: 0 0 0 0"`, `got "shl 64: 1 2 -1 -68719476736"`.
2. **Made `shift_right` logical** → 3 unit tests red, 6 golden lines diverged: `want "shr arithmetic: -4 …"`, `got "shr arithmetic: 9223372036854775804 …"`.
3. **`bitwise_xor` computing `|`** → golden red on the WebSocket round trip: `want "ws mask: 222 22 PASS"`, `got "ws mask: 222 222 FAIL"`. **The Rust unit tests stayed green** — they only cover the shift helpers, so and/or/xor/not are covered by the `.scrl` program alone.
4. **A bridge-only bug** (`opc::BIT_XOR => self.bit_or()`) → the native test red, the interpreter golden **green**. That asymmetry is why the JIT test exists: a golden program cannot see a bridge bug.

The JIT path is exercised rather than merely wired — `SCARLET_NATIVE_DEBUG=1` shows `warmed fn#223 spin`, `fn#224 shifty`, `fn#225 masky`, all three identity loops returning correctly.

## Known

- **No hex literals.** `println(0xFF)` gives `error: Expected ')', got 'xFF'`. Every mask in the tests and docs is decimal. Filed separately rather than folded in.
- No logical (unsigned) right shift and no operators. Both are additive later; the docs tell callers to mask.
- `shift_right(x, i64::MIN)` returns 0 for negative `x` rather than `-1` — it delegates left by a saturated `i64::MAX`, landing past the width. Total and documented, and it is the one corner where "negative count reflects" and "past the width keeps the sign" interact.

## No timing number

A fleet was live on the host throughout. The XOR argument is made from operation count in the ticket: emulating XOR per byte in arithmetic costs roughly 8 div/mod pairs per byte, so a 4-64 KB WebSocket frame becomes ~10^5-10^6 extra operations per frame in each direction.


alii:
Adversarial review returned **do not merge** on two blockers. Both fixed at `a061d0d`, each verified by re-running the reviewer's own experiment.

**Blocker 1 — the bridge test passed with a bridge arm mis-wired.** Reproduced first: with `opc::BIT_OR => self.bit_xor()` planted, the old test printed `ok`. Cause was identity operands — the only OR operand was `0`, the shared identity of OR and XOR, and the AND arm was caught *only* because the natively-executed iteration count happened to be odd on that host.

Replaced with two state-evolving loops that fold the counter every step, so there is no fixed point to collapse onto. `mixer`'s OR bits are a **strict subset** of its AND mask, which is what makes OR's contribution survive the AND and reach the output — that subset relation is the specific fix for the uncovered arm. All six arms now caught:

| plant | caught |
|---|---|
| `BIT_OR => bit_xor` | yes (was **no**) |
| `BIT_AND => bit_xor` | yes, by construction |
| `BIT_XOR => bit_or`, `BIT_NOT => neg`, `BIT_SHL => bit_shr`, `BIT_SHR => bit_shl` | yes |

Parity check: with `BIT_AND => bit_xor` planted, moving `WARM_CALLS` 8 → 9 still fails, with *different* wrong values. The parity flips and the test catches it on both sides.

Expected output is recomputed in Rust from the same recurrence, and cross-checked against an independent Python implementation before either was trusted.

**Blocker 2 — no witness that native code ran.** The test now runs under `SCARLET_NATIVE_DEBUG=1` and asserts a witness keyed on the **function name** — reads `selected fn#N <name>` for each loop's index, then requires a matching `warmed fn#N`. Re-ran the reviewer's experiment with `WARM_CALLS = 4_000_000_000`: **now FAILED**, where before it passed with the JIT switched off entirely. stdout is *correct* in that run, since the interpreter gives the same answer — only the witness catches it.

**Two doc laws corrected, with their domains.** The negative-count identity holds for every `n` except `min_value`, where `0 - min_value` wraps and both sides land past the width in opposite directions — neither value is wrong, the identity between them stops there. The logical-shift recipe is replaced by shift-then-mask for `n >= 1`, verified exhaustively: 2112 pairs against a real `u64` shift, 0 mismatches for `n >= 1` and 9 at `n == 0`, which is why the domain starts at 1.

Two new unit tests, both watched failing — including one planted with the *old doc recipe*, which is the bug that test exists to catch.

The same wrong mask-first recipe appeared in `exec.rs` and is fixed there too: correcting a falsehood and leaving its twin next door is not a fix.

Gates: `fmt --check` 0, `clippy --all-targets` 0, `test --workspace` 0. Both headline plants re-run against the final tree, then restored and rebuilt, so nothing rests on a stale binary.

PR #29, opened 2026-08-12
carry a variant's declared name on its constructor scheme

alii:
An aliased constructor import (`import ./color.{Green as G}`) never matched when the match also had a catch-all arm. No error, no warning — the arm was dead and control took the wrong branch. It shipped, and it broke RESP3 pub/sub in the redis client.

## It is a value-identity bug, not a matching bug

The ticket framed this as patterns failing to match. The cause is wider:

```
println(G(9))                     was  G(9)     now  Green(9)
println(G(9) == color.Green(9))   was  False    now  True
```

**Two values of the same variant with the same payload compared unequal** because one was built through an alias.

## Root cause

`resolve.rs`'s `from_kind` built a `VariantRef` taking three fields from the declaration and `variant_name` from **the binding** — which an aliased import makes `G`. `ValueKind::Constructor` never carried the declared variant name, despite a comment claiming it "carries enough to compile pattern-match and constructor-call".

The catch-all only decides *which consumer notices*. `emit_ladder` tests `Op::MatchEnum` against the name **string**; `emit_switch`, taken when the match is exhaustive by heads, dispatches on the tag and is immune. Confirmed in the disassembly: `MatchEnum` against `PushConst "G"` while the value carried `Green`.

**Fix:** `ValueKind::Constructor` carries `variant_name` from the declaration, and `from_kind` **takes no name parameter at all**, so a use site's spelling cannot reach codegen.

## Why the previous fix missed it

All four tests in `9999018` build their scrutinee **in the matching file** and omit a catch-all. Both halves mattered, and the second is the one nobody spotted: a value built through the alias carried the *alias* name, so an alias-named test matched it by accident. **The four tests were self-consistent in exactly the way real code is not.**

## The failing tests came first

Eight tests written against pristine `d49ebb8` before any compiler change:

```
test result: FAILED. 9 passed; 6 failed
an_aliased_arm_matches_beside_a_catch_all ... FAILED
an_aliased_arm_matches_beside_a_bare_binding_catch_all ... FAILED
an_aliased_arm_matches_a_qualified_scrutinee_beside_a_catch_all ... FAILED
an_aliased_head_matches_nested_beside_a_catch_all ... FAILED
an_aliased_head_matches_in_an_or_pattern_beside_a_catch_all ... FAILED
an_alias_does_not_change_a_constructed_value_identity ... FAILED
```

Reverting only the six compiler files afterwards, keeping the tests, reproduces the same six.

## Neighbouring shapes

| shape | before | after |
|---|---|---|
| nested, in a tuple, + catch-all | fell through | fixed |
| or-pattern + catch-all | fell through | fixed |
| bare binding `other ->` | fell through | fixed |
| `let` destructure, value from declaring module | correct | correct |
| catch-all **before** the aliased arm | rejected unreachable | rejected |

The last two were already correct and act as guards against the fix rather than reproductions. The destructure survived because an irrefutable single-variant unwrap emits no name test.

## Gates

`fmt` 0, `clippy` 0, `test` 0 — 41 `test result: ok`, zero failures. `SCARLET_GC_STRESS=1` likewise 0. `dylint` 0.

**`cargo hawk check -D warnings` is UNRUN, not passed** — `error: no such command: hawk`. It is a CI gate and this host does not have it.

## Filed, not fixed

- `emit_ladder` identifies variants by name string while `emit_switch` uses the tag. The source now makes all sites agree; their *ability* to disagree is untouched, and the same name-based sites exist in the JIT. Hot path, wants its own measurement.
- The docs say selective imports pull "types, constants, and functions". Constructors are missing from that list. Pre-existing.

This does not close the two open `check_parity` jump-over tickets — it touches pattern head resolution, not jump patching. `check_parity` passes.

### DIFF
--- a/crates/scarlet/build.rs
+++ b/crates/scarlet/build.rs
@@ -354,10 +354,11 @@ fn valuekind(k: ValueKind) -> String {
             type_name,
             type_id,
             variant_idx,
+            variant_name,
             arity,
             field_labels,
         } => format!(
-            "ValueKind::Constructor {{ type_name: {type_name:?}, type_id: {}, variant_idx: {variant_idx}, arity: {arity}, field_labels: {} }}",
+            "ValueKind::Constructor {{ type_name: {type_name:?}, type_id: {}, variant_idx: {variant_idx}, variant_name: {variant_name:?}, arity: {arity}, field_labels: {} }}",
             tid(type_id),
             aslice(field_labels)
         ),
--- a/crates/scarlet/tests/modules.rs
+++ b/crates/scarlet/tests/modules.rs
@@ -579,6 +579,44 @@ fn an_alias_does_not_change_a_constructed_value_identity() {
     run_project_outputs(&proj, "run", "main.scrl", "Green(9)\nTrue\nTrue\n");
 }
 
+/// An alias is free to collide with a real variant of the same type, and the
+/// collision is where a wrong name stops being a dead arm. `Green as Red`
+/// made the ladder test for `"Red"`, which `color.Red` carries: the arm was
+/// entered against a nullary variant, and binding its field 0 indexed past
+/// the end of the value stack — an interpreter panic, exit 101, after two
+/// wrong lines of output.
+///
+/// The tests above all alias to a fresh name, where the mismatch only ever
+/// costs a branch. None of them can reach this.
+#[test]
+fn an_alias_colliding_with_a_real_variant_does_not_capture_it() {
+    let proj = Project::new("alias_ctor_collide");
+    proj.write("color.scrl", COLOR_MAKE_SRC);
+    proj.write(
+        "main.scrl",
+        "import ./color\nimport ./color.{Green as Red}\n\nprintln(Red(5))\nmatch color.make(5) {\n\tRed(_s) -> println(1)\n\t_ -> println(0)\n}\nmatch color.Red {\n\tRed(_s) -> println(1)\n\t_ -> println(0)\n}\n",
+    );
+    run_project_outputs(&proj, "run", "main.scrl", "Green(5)\n1\n0\n");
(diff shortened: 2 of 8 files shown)


alii:
Adversarial review returned no blockers. Three tests added at `bafdba9`, each watched failing against master's compiler files:

- `an_alias_colliding_with_a_real_variant_does_not_capture_it` — `{Green as Red}` where `Red` is a real variant is not a wrong branch on master, it is an interpreter panic: `index out of bounds: the len is 368 but the index is 368` at `exec.rs:560`. Nothing pinned the most severe manifestation.
- `a_warmed_aliased_match_agrees_across_the_warmup_boundary` — master returns `19993` where the answer is `20000`. Exactly seven iterations wrong out of 20,000, because `WARM_CALLS` is 8: the divergence is the calls before the body compiles. A 0.035% wrong answer inside one process, clean exit, empty stderr. No cold test could see it, and every test in this PR ran cold. The `warmed fn` assertion is load-bearing — without it a green could mean the body never left the interpreter.
- `an_aliased_stdlib_constructor_reaches_a_vm_builtin` — the sole exercise of the changed `build.rs` line. Master: `binary.from_int_ascii: expected Radix, got 'Radix'`.

The review also corrected this PR's own follow-up ticket: `clif.rs:2469/2536` are constructor sites, not comparisons. The JIT dispatches on a packed tag and does no name comparison anywhere — **the JIT was already correct and the interpreter was the wrong engine.** Recorded there.

Gates: `cargo test --workspace` 0, `clippy --all-targets` 0, `fmt --check` 0. `cargo hawk check` is UNRUN — not installed on this host. `SCARLET_GC_STRESS` is set by CI and read nowhere in the tree, so it is a no-op, not a gate.

PR #32, opened 2026-08-12
decode: bind-shaped decoders that still accumulate, with no placeholder

alii:
The `map2`…`map5` ladder stops at 5 and `map3` is literally `map2(map2(…))`, so a 40-field object could not be written at all. This replaces it with backpassing, keeps multi-error accumulation, and **never asks the caller for a placeholder value**.

## The design question, and where the wall actually is

The trilemma is {bind ergonomics, accumulation past a failure, no witness of the decoded type} — a library gets two. Backpassing desugars `x <- f(d)` to `f(d, fn(x) rest)`; the continuation is opaque, so by parametricity the only way to reach the `b` in `fn(a) Decoder(b)` is to apply it to an `a`. When the step failed there is no `a`, and Scarlet is strict with no bottom. Witness, or short-circuit.

**But the objection to Gleam is not the witness — it is being asked for one, and that distinction is the whole design.**

Gleam's `Decoder(a)` is `fn(Dynamic) #(a, List(DecodeError))`: an **unconditional** value channel, so every decoder must always produce an `a`, and `decode.failure(zero, expected)` is the bill. Scarlet's is `Result(a, Array(Failure))`, and `Err` inhabits that for every `a` — which is exactly why `fail` needs nothing here. **The zero is the price of an unconditional value channel, not the price of bind.**

So keep the conditional channel and move the recovery witness onto the *decoder*, where the library writes it once:

```scarlet
type Recovery(a) { Resume(value a) | Halt }   // private — no public API mentions it
```

Primitives carry their own; every combinator derives one from its parts; `optional_field` answers `absent` always, because `Field(a)` is inhabited without an `a`. `also` runs the continuation against the witness to collect its failures, then **still returns `Err`** — a recovered value can never reach an answer.

## Before / after

5 members went from a `map5` call with a 5-argument lambda to five `<-` lines. **40 members could not be written before**; after, it is 40 lines with no nesting — real and running in `tests/programs/decoders.scrl`, and with 2 of the 40 malformed the golden reports both, with the 38 between them still checked.

## Given up, stated rather than hidden

1. A decoder built from `fail` answers `Halt` and stops accumulating there. `guard(dec, ok, expected)` keeps its type and so keeps its recovery — `guard` reports 2 where `then+fail` reports 1.
2. `then` never accumulates. That is correct rather than conceded: failures past a bad discriminant belong to a branch the payload never selected.

## Type-system improvements, each proved by a compiler error on the old spelling

- `one_of(first, rest)` is non-empty by construction — `one_of([])` was a *runtime* failure describing the decoder rather than the document.
- `Failure.found` is a `Found`, not a `String` — `'no such member'` was a magic string indistinguishable from a value that reads that way.
- `fail` takes no value at any type: `decode.fail('a user', User(name: ''))` → `Expected 1 argument(s), got 2`.

## Gates

`fmt` 0 · `clippy` 0 · `test --workspace` 0 (42 binaries, 1265 passed) · `gen-editor-syntax --check` 0

**No goldens moved** — `json.stdout` is not in the diff, so no type id shifted.

Follow-ups filed: **T-212** (`wire.scrl`'s five helpers and 40 call sites), **T-211** (`ApplicativeDo` in `desugar.rs` — the compiler can see whether a backpass continuation mentions its binder, and lowering the independent case to `map2` would delete `Recovery` outright; the wall is a library wall, not a language one).

### DIFF
--- a/crates/scarlet/tests/golden/programs/decoders.stdout
+++ b/crates/scarlet/tests/golden/programs/decoders.stdout
@@ -1,0 +1,38 @@
+-- a forty-member record, one line per member --
+40 members    10 hi mentions=2 guild=present(13) nick=null ref=absent
+40, two wrong 2: .id: expected a quoted integer, found an integer
+.tts: expected a boolean, found a string
+
+-- independent members accumulate, past the old five-member ceiling --
+8 members ok  Row{ a: 1, b: x, c: True, d: 2, e: y, f: False, g: 3, h: z }
+3 of 8 wrong  3: .a: expected an integer, found a string
+.e: expected a string, found an integer
+.h: expected a string, found a boolean
+only 1st wrong1: .a: expected an integer, found a string
+
+-- absent, null and present stay three outcomes --
+absent        absent -> Some(held)
+null          null -> None
+present       present(ada) -> Some(ada)
+bad id + null 1: .id: expected an integer, found a string
+bad optional  1: .nick: expected a string, found an integer
+
+-- one_of: the same field written 2 or 2.0 --
+2 and 2.0     Pair{ v: 2, w: 2 }
+2.0 and 2     Pair{ v: 2, w: 2 }
+neither       2: .v: expected an integer, found a string
+.v: expected a number, found a string
+
+-- fail takes no value, at any type --
+fail String   1: <root>: expected a string this program approves of, found an object
+fail Message  1: <root>: expected a message this program approves of, found an object
+halts after   1: <root>: expected a string this program approves of, found an object
+
+-- guard refines without halting; then+fail halts --
+guard, both   2: .v: expected a positive integer, found an integer
+.w: expected a positive integer, found an integer
+then+fail     1: .v: expected a positive integer, found an integer
+
+-- a bad discriminant reports itself, not an untaken branch --
+chat          Chat(hi)
+bad tag       1: .t: expected a string, found an integer
--- a/crates/scarlet/tests/golden_examples.rs
+++ b/crates/scarlet/tests/golden_examples.rs
@@ -271,6 +271,13 @@ suite! {
         // adversarial answers: 1e400, a lone surrogate, invalid UTF-8, the
         // 64-bit boundary, duplicate keys and deep nesting.
         json,
+        // scarlet/json/decode's own shape: a forty-member record written one
+        // flat line per member, independent members accumulating their failures
+        // past the old five-member ceiling, three-way presence under that
+        // accumulation, one_of over 2 vs 2.0, and the two places where
+        // accumulation is deliberately given up — `fail`, and `then`'s
+        // dependent continuation.
+        decoders,
     ],
 
     // No golden, because there is no fixed output: `http_server` ends in an
--- a/crates/scarlet/tests/programs/decoders.scrl
+++ b/crates/scarlet/tests/programs/decoders.scrl
@@ -1,0 +1,417 @@
+import scarlet/array
+import scarlet/float
+import scarlet/json
+import scarlet/json/decode
+import scarlet/json/field
+
+// The decoder API at the scale it exists for.
+//
+// `json.scrl` covers the parser and the presence states. This file covers the
+// shape of `scarlet/json/decode` itself: that a record decoder is one flat
+// line per member with no arity ceiling, that independent members accumulate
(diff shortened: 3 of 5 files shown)


alii:
## Adversarial review: DO-NOT-MERGE

**A one-character type error in an untrusted JSON payload hangs the process.**

`also` runs the continuation against a manufactured value — a `0`/`''`/`[]` the document never contained. Any continuation whose *termination* depends on the value it was handed now loops forever on input the decoder is in the middle of rejecting. This is the Discord gateway; the input is hostile by construction.

No `guard`, no exotic combinator — plain `int()` and a quoted number:

```scarlet
rows <- decode.take('rows', decode.list(decode.int()))
per  <- decode.take('per',  decode.int())
decode.succeed(Page(rows: rows, per: array.length(chunk(rows, per))))
```
`{"rows":[1,2,3,4],"per":2}` → ok. `{"rows":[1,2,3,4],"per":"2"}` → **exit 124**, no output, no failure ever reported. `chunk` terminates for every `per >= 1`; `also` hands it `0`.

**Boundary experiment:** the same program with `then`/`field` instead of `take` — same branch, same binary — prints `err .per: expected an integer, found an integer` and exits 0. `also`'s speculative execution is the whole of the difference.

### It is not "twice", it is "twice with different values"

A nested record runs the continuation twice, and the second pass sees fabricated values for fields the document set **correctly**: `q=2` on the first pass, `q=0` on the second. Effects are trivially reachable — `println` is an ordinary function returning `Nil`.

And the destructive case the ticket exists to prevent is reachable through that door. A PATCH applier written as an effect emits `nick: CLEARED` on good input, then `nick: unchanged` from a *recovered* `absent` when an unrelated field is malformed. **A recovered `absent` is indistinguishable from a decoded one at the point of use.** The header's claim that "a recovered value never reaches the caller" is true of the value channel and false of everything else.

### The error channel reports failures that are not in the document

`k` has type `fn(a) Decoder(b)` — *identical* to `then`'s. Nothing distinguishes an independent member from a dependent one, so a dependent continuation is built from the fabricated value and its failures are appended as the document's. On `{"kind":7,"radius":3}` it reports `.side: expected a member named "side", found no such member`. There is no `.side` in that document and never was. The module argues at length that this is exactly why `then` must short-circuit — and `also`, the combinator it is shaped around, does the thing it condemns.

### Two more, both undocumented

- **`guard`'s recovery does not satisfy `guard`'s predicate.** `guard(int(), fn(n) n > 0, …)` has recovery `Resume(0)`. The one combinator whose job is "values satisfying `ok`" hands the continuation one that does not — which is what makes the hang reachable even in code that defended itself.
- **`one_of`'s accumulation depends on argument order.** Only the first alternative's recovery is reachable: same alternatives, same document, 2 failures one way and 3 the other.

### The suite cannot fail for any of this

Not vacuous — three planted defects all went red. But every continuation in it is pure and value-independent, every record flat, and there is **no nested-record failure case anywhere**. So it exercises none of: an effectful decoder (the author's own stated worry), a value-dependent continuation, a nested record on a failure path, or a recovered `absent` reaching a continuation that acts on it. The four things the design is risky about are the four it does not test.

## Verified clean

Gates all 0 (fmt, clippy, gen-editor-syntax, `test --workspace`). **The no-goldens-moved claim is true and independently checked** — `json.stdout` untouched despite being rewritten from `map4` to `take`; nothing regenerated to hide a shift. The accumulation headline is real and was reproduced rather than taken. Three-way presence is genuinely witnessed. The value-channel safety property is genuinely witnessed — a recovered value escaping as an answer *would* be caught. `Halt`'s concession is exactly the stated size. The thunk is genuinely not forced on the success path.

Manufactured values cannot index or divide out of bounds **in today's stdlib** — `10 / 0` returns `0` and `scarlet/array` exposes no partial indexing. That defuses the crash version of finding 1, but only by the current stdlib surface, not by anything in this design. A partial `array.get` or a trapping division turns the hang into a crash.

## The shape of the fix

`also` needs the manufactured value confined to something that cannot reach user code — or the continuation needs a type that says it does not depend on its argument. The latter is **T-211**: the compiler can already see whether a backpass continuation mentions its binder, and lowering the independent case to `map2` deletes `Recovery` outright. The author called the wall a library wall rather than a language one, and this review is the evidence for that.

PR #43, opened 2026-09-20
drop dangling mod lines

alii:
First PR in the stack on #42.

#42 deleted files, but the `mod` lines that pointed at them stayed. With those lines left in, `rustfmt` can't load the crates at all, so `cargo fmt --all --check` fails before it checks any formatting. This removes the 12 dangling lines:

- `scarlet`: `dis`, `vm`
- `scarlet_core`: `precompile`, `static_ir`, `core_ir::{clif, codec, emit, native_frame}`, `bytecode::{binop, peephole}`, `bytecode::compiler::abi`, `typed_ir::wire`

`lsp/mod.rs`'s `mod wire;` stays: it's the LSP's own file and still exists.

## Checks

- `cargo fmt --all --check`: passes now (it failed on #42)
- grammar (install, generate + diff, corpus): passes
- `cargo check --workspace`: still fails first at `scarlet_types`, on its two ties to the old VM. That's expected on this stack, and the next PR starts cutting them.
- clippy, test, hawk, dylint: not run. They need the workspace to compile, which it won't until the new VM exists.


alii:
Squashed into `rewrite` as ae63b30 (`drop dangling mod lines`), one commit per PR. It lands on master with #42.

PR #44, opened 2026-09-20
types owns TypeId

alii:
Stacked on #43.

`TypeId` is the id of a user-declared type. The front end creates it once per `type` declaration (`TypeEnv::register_type_head`), and the runtime only carries it around in a value's header. It used to be defined in `scarlet_vm` and re-exported by `scarlet_types`. This PR moves the definition into `scarlet_types/src/type_def/mod.rs`, unchanged: same `i32` layout, same `NONE` sentinel, same `Display`.

When the new VM exists, it will carry its own runtime tag, and the compiler will turn a `TypeId` into that tag. The front end no longer needs the VM for this.

## Checks

- `cargo check -p scarlet_types`: down from 2 errors to 1. The one left is `Builtin { op: scarlet_vm::bytecode::Op }` at `crates/scarlet_types/src/types/infer.rs:275`, which gets its own PR.
- clippy, test, hawk, dylint: not run. The workspace doesn't compile yet (see #42).


alii:
Squashed into `rewrite` as b5b4883 (`types owns type ids`), one commit per PR. It lands on master with #42.

PR #45, opened 2026-09-20
types name builtins by intrinsic

alii:
Stacked on #44. This is option B from the review thread: a list of the built-ins, instead of the old VM's opcode.

## What

- New `crates/scarlet_types/src/intrinsic.rs`. It defines an `Intrinsic` enum with one variant per `@vm(key)` in the stdlib: 135 in all, taken from the old `builtin_op` table in `bytecode/mod.rs`, which matches the stdlib exactly. For example, `@vm(string__length)` is `Intrinsic::StringLength`.
- `Intrinsic::from_key("string__length")` gives `Some(Intrinsic::StringLength)`, and an unknown key gives `None`. A single `intrinsics!` macro writes both the enum and the key mapping, so the two can't drift apart.
- `ValueKind::Builtin { op: scarlet_vm::bytecode::Op }` becomes `ValueKind::Builtin { intrinsic: Intrinsic }`.
- The crate doc now says what's true: `scarlet_types` depends only on `scarlet_syntax`.

## Why

The name gets turned into an `Intrinsic` once, at the `@vm(...)` line, and everything after carries the variant. There's no second lookup by string later that "can't fail". What each intrinsic *does* belongs to the runtime; the types crate only names them.

Nothing creates an `Intrinsic` yet. The next PR does that in `scarlet_core`'s analysis, where `builtin_op(name)` is called today.

## Checks

Scoped to the crates that compile now (`scarlet_syntax`, `scarlet_types`):

- `cargo clippy -p scarlet_syntax -p scarlet_types --all-targets -- -D warnings`: passes
- `cargo test -p scarlet_syntax -p scarlet_types`: 261 passed, including the two new tests. One checks that every key names its own variant; the other that an unknown key gives nothing.
- `cargo dylint --all -- -p scarlet_syntax -p scarlet_types` with `DYLINT_RUSTFLAGS=-D warnings`: passes
- Workspace-wide clippy, test and hawk: not run. `scarlet_core` doesn't compile yet (see #42).


alii:
Squashed into `rewrite` as 9b381a1 (`types name builtins by intrinsic`), one commit per PR. It lands on master with #42.

PR #46, opened 2026-09-20
analysis mints intrinsics

alii:
Stacked on #45. This is where an `Intrinsic` gets made.

## What

- `bytecode/analysis.rs`: when a stdlib function has `@vm(key)` instead of a body, analysis now calls `Intrinsic::from_key(key)`. It used to call `builtin_op(key)`. The scheme gets `ValueKind::Builtin { intrinsic }`. An unknown key is still the same compile error, `unknown @vm builtin '…'`, at the annotation.
- `bytecode/mod.rs`: the old `builtin_op` table (135 lines of `"key" => Op::…`) is deleted. The list now lives in one place, `scarlet_types::intrinsic`.

## What it doesn't do yet

The typed IR still carries `TypedCallee::Builtin { op: Op, imm: Imm }`, and so do `typed_ir/resolve.rs` and `core_ir/lower.rs`. Those move to `Intrinsic` in their own PRs.

## Checks

- `cargo check -p scarlet_core`: down from 222 errors to 85, with none left in `analysis.rs`
- clippy, test, hawk, dylint on `scarlet_core`: not run, because it doesn't compile yet (see #42)


alii:
Squashed into `rewrite` as 50de2d5 (`analysis mints intrinsics`), one commit per PR. It lands on master with #42.

PR #47, opened 2026-09-20
core ir owns its ids

alii:
Stacked on #46.

The core IR numbers things with small typed ids: `LocalId` (a local), `JoinId` (a jump target), `ConstId` (a constant) and `FuncIdx` (a function). Each is a `u32` in its own type, so they can't be mixed up. The macro that declares them (`newtype_index!`) and the vector only they can index (`TiVec`) used to live in `scarlet_vm`.

## What

- `crates/scarlet_core/src/tivec.rs` is the old `scarlet_vm/src/tivec.rs`, moved as-is, tests included. The only change is one doc line that listed "code addresses".
- `core_ir/mod.rs` declares `FuncIdx` itself (same `"fn#"` display). It uses `crate::newtype_index`, not `scarlet_vm`'s.
- Removed `CodeAddr`: it was a bytecode jump offset that only the deleted emitter used.
- Removed the `native_rc` re-export: it was JIT support, and nothing here uses it.

## Checks

- `cargo check -p scarlet_core`: down from 85 errors to 39, with none left about ids
- The moved `tivec` tests can't run until `scarlet_core` compiles


alii:
Squashed into `rewrite` as 50f78ab (`core ir owns its ids`), one commit per PR. It lands on master with #42.

PR #48, opened 2026-09-20
compile the stdlib from source

alii:
Stacked on #47.

The old build baked the stdlib into the binary as old-VM values (`build.rs` → `precompile.rs` → `static_ir`). At startup the compiler copied that snapshot in (`seed_static`), instead of compiling the stdlib's `.scrl` files. #42 deleted the baking side. This removes the loading side, so the stdlib is always compiled from its embedded source. That path already existed: the LSP used it when you edit the stdlib itself.

## What

- `compile`, `check` and `compile_with_native` drop their `pre: Option<&'static StaticStdlib>` argument. `CompileOptions` drops its `stdlib` field, and `CompileOptions::new(base_dir)` takes only the base directory.
- `compile_with` always calls `register_prelude()`.
- Removed from `Compiler`:
  - `seed_static`
  - the `static_stdlib` field
  - `CompilerParts`, and the accessors only `precompile.rs` used (`into_parts`, `take_module_table`, `take_type_info`, `take_frame_layouts`, `set_native_hook`)
  - `is_reserved`'s second lookup, into the baked sorted list
- `ModuleTable` loses `static_fallback` and `get_or_hydrate`. The three callers use plain `get`.
- `IncrementalSession::new()` takes no argument and compiles the embedded stdlib. It shares one constructor with `new_from_source(root)`, and gains a `Default` impl. That shared constructor also lets the prelude-teardown rebuild drop an `expect`.
- Comments that described the baked blob now say what's true.

## Worth knowing

- Startup now compiles the stdlib every run. The old comment put that at about 9 ms. The new VM can bring back a cache later if it's worth it.
- A few leftovers are now unused, and get their own PR next: the `Hydrated` module origin, `ModuleTable::into_loaded`, and `InferEngine::seed_arena` / `ArenaSeed` / `TypeEnv::take_type_info` in `scarlet_types`.

## Checks

- `cargo check -p scarlet_core`: down from 39 errors to 26, with none about the baked stdlib


alii:
Squashed into `rewrite` as 784245a (`compile the stdlib from source`), one commit per PR. It lands on master with #42.

PR #49, opened 2026-09-20
drop hydrated modules

alii:
Stacked on #48.

With the baked stdlib gone (#48), nothing creates a "hydrated" module any more. Every loaded module is now compiled from source, either embedded or on disk. This removes the code that existed only for hydrated ones.

## What

- `module/mod.rs`:
  - `ModuleOrigin::Hydrated` is gone, along with `CachedModule::hydrated`, the test-only `insert_hydrated`, and `into_loaded`.
  - `CachedModule::module_refs()` always has refs now, so it returns `&Rc<ModuleReferences>` instead of an `Option`.
  - The two tests that inserted a hydrated module now insert an `Embedded` one.
- `bytecode/session.rs`: the reference-graph builder used to invent definitions for hydrated modules from their interface (`SynthDef`, `synth_refs_from_interface`). Every cached module has real references now, so the builder just inserts them. That's about 100 lines gone, plus four imports.
- `scarlet_types`: `InferEngine::seed_arena`, `ArenaSeed` and `TypeEnv::take_type_info` go too. They existed only to copy the baked snapshot in and out.

## Checks

- `cargo clippy -p scarlet_syntax -p scarlet_types --all-targets -- -D warnings`: passes
- `cargo test -p scarlet_syntax -p scarlet_types`: 261 passed
- `cargo check -p scarlet_core`: still 26 errors (this PR only removes unused code), with none in `module/` or `session.rs`


alii:
Squashed into `rewrite` as 55fbe90 (`drop hydrated modules`), one commit per PR. It lands on master with #42.

PR #50, opened 2026-09-20
builtins travel as intrinsics

alii:
Stacked on #49.

#46 made analysis create an `Intrinsic` for each `@vm` function. This carries that `Intrinsic` the rest of the way, so a built-in call never turns back into an old-VM opcode:

- `ValueKind::Builtin { intrinsic }`, then `Den::Builtin { intrinsic }` and `EtaTarget::Builtin { intrinsic }` (`typed_ir/resolve.rs`)
- then `TypedCallee::Builtin { intrinsic }` (`typed_ir/mod.rs`)
- then the new core-IR atom `Atom::Intrinsic { intrinsic, args }` (`core_ir/mod.rs`, `core_ir/lower.rs`)

Other basic operations (`Atom::PrimOp`, like add-int or tuple-index) still use old opcodes. They get their own PR.

## `wire` loses its descriptors, for now

`wire.encode`/`wire.decode` were the only built-ins that carried an extra operand. The elaborator built a descriptor of the type being encoded and attached it to the old VM instruction as an `Imm::WireDesc` index. With no instruction to attach it to, this removes that plumbing:

- the elaborator's `wire_imm` / `eta_wire_imm`, and `ElabCtx::wire_descriptor`
- `Compiler::wire_descs`, the `WireCtx` impl and the `wire_builtin` / `wire_handle` helpers (`bridges.rs`)
- `Imm::WireDesc`, and the `imm` argument of `eta_wrapper`
- the tests for descriptors and wire templates

One behaviour change: `scarlet check` no longer rejects a `wire.encode` of a type that can't cross the wire, such as one holding a closure. That check belonged to the descriptor builder. It should come back with the new VM's wire design, which also has the audit's findings to fix (forged opaque types, tampered closures).

## Checks

- `cargo check -p scarlet_core`: down from 26 errors to 22, with none left in `resolve.rs`, `eta.rs` or `elaborate.rs`'s wire code


alii:
Squashed into `rewrite` as be2b4e2 (`builtins travel as intrinsics`), one commit per PR. It lands on master with #42.

PR #51, opened 2026-09-20
loads nil and bools atoms

alii:
Stacked on #50. This is the first of three PRs that take the old VM's opcodes out of the core IR.

Reading a variable, `Nil`, and the `true`/`false` that `&&`/`||` produce were all spelled as stack-machine instructions in the core IR. For example, a module-scope read was `Atom::PrimOp { op: Op::PushGlobal, imm: Imm::Index(3) }`. Now they are atoms of their own:

- `Atom::Load(Load)`, where `Load` is one of:
  - `Slot(FrameSlot)`: a selective-import slot
  - `Global(GlobalSlot)`: a module-scope binding
  - `Capture(CaptureIdx)`: a captured value
  - `SelfClosure`: the current closure

  They carry the typed IR's own typed indices, not a `u16` narrowed for an instruction operand.
- `Atom::Nil`, for an empty block.
- `Atom::Bool(bool)`, for the `false` in `a && b` → `if a { b } else { false }`, and the `true` in `||`.

They print as `global3`, `nil`, `true` and so on. The typed IR's `ValueRef` docs stop describing each place by the opcode that used to read it.

## Checks

- `cargo check -p scarlet_core`: still 22 errors. This PR adds no fixes and no new errors; the remaining ones are the next PRs.


alii:
Squashed into `rewrite` as 4eb5165 (`loads nil and bools atoms`), one commit per PR. It lands on master with #42.

PR #52, opened 2026-09-20
core ir names its prim ops

alii:
Stacked on #51. This is the second and last PR that takes the old VM's opcodes out of the IRs.

## What

**`core_ir/prim.rs`** is new: the `PrimOp` enum. It lists every basic operation the core IR uses, each with a one-line doc saying what it does and what its arguments are. It's the file to read first. The groups:

- Int, Float and String arithmetic and comparison. For example, `IntAdd`, `FloatLt` and `StringConcat`.
- Structural `Eq`/`Ne`, and `Not`.
- **Unresolved** forms (`Add`, `Lt`, …). These are for an operand whose type is still a constrained variable, like `fn add(a, b) { a + b }`, so the backend picks the operation at run time. This keeps today's behaviour.
- Tuples and records: `MakeTuple`, `TupleField(n)`, `Field(n)`, `FieldUnchecked(n)`.
- Arrays: `MakeArray`, `ArrayIndex`, `ArrayIndexOr`, `ArraySlice`, `ArrayElem(n)`, `ArrayDrop`, and so on.
- Binaries: `BinaryFromInt`, `BinaryView`, `BinaryReadUtf8`, and so on.

**`Imm` is gone.** Whatever was fixed at compile time moved into the variant: `TupleField(2)`, not `TupleIndex` plus `Imm::Index(2)`. An argument count is just `args.len()`. `xs[i] or d` always passes `d` as a third argument, instead of switching between "constant in the operand" and "pushed on the stack"; the fused form already only applies when `d` is pure, so evaluating it up front changes nothing.

**`typed_ir/binop.rs`** is new. It replaces the `bytecode/binop.rs` that #42 deleted: the elaborator used it to pick a specialised operation for `+`, `<` and the rest. It's the same logic, now returning a `PrimOp`. `TypedExpr::Binary` and `TypedExpr::Unary` carry a `PrimOp`, and so does `Atom::PrimOp`.

**Pattern-matching arithmetic** (array-length checks, range bounds, binary cursor maths) now says what it is, `IntAdd`/`IntGe` and so on, instead of the generic `Add`/`Gte`. The operands were always Ints.

The name is `PrimOp`, not `Prim`, because `types::Prim` already means a primitive *type* (Int, Float and so on).

## Checks

- `cargo check -p scarlet_core`: down from 22 errors to 19, with no opcode errors left in `typed_ir/` or `core_ir/`. What's left is the bytecode emitter's glue, the JIT hook, and `Value` in the constant pools, each its own PR.
- The new `binop` tests can't run until `scarlet_core` compiles


alii:
Squashed into `rewrite` as 82b56ab (`core ir names prim ops`), one commit per PR. It lands on master with #42.

PR #53, opened 2026-09-20
constants are front-end values

alii:
Stacked on #52.

Literals (`42`, `1.5`, `'hi'`, `<<1, 2>>`) and the constants the elaborator needs with no literal behind them (an array pattern's length, a segment's width) were stored as old-VM `Value`s. The pool itself was the old VM's `Program.constants`, built through its frozen heap. Now the front end owns them.

## What

- New `core_ir::Const`: `Int(i64)`, `Float(f64)`, `String`, and `Binary { bytes, bit_len }`. It sits next to `ConstId`, which indexes it.
- Equality and hashing go through a small key type (`ConstKey`), and a float compares by its bits. So `0.0` and `-0.0` stay two constants, and a constant always equals itself. That follows the repo's rule that a map key is the canonical identity of what it names. The old dedup keyed on `Value::to_bits()`, which the repo's own notes list as a past bug, because for a boxed value those bits are a pointer.
- The `Compiler` gets its own `consts: Vec<Const>` and `const_dedup: HashMap<Const, ConstId>`. `add_constant`, `const_int`, `const_str` and `const_binary` return a `ConstId`, so there are no more `as i32` / `as u32` round trips.
- `number_literal_value` and `NumLitError::recovery` build a `Const` directly. The check walk types a number literal with `matches!(c, Const::Float(_))`.
- `TypedProgram.consts` and `CoreProgram.consts` are `Vec<Const>`. `CoreProgram::default()` no longer needs a pooled nil, now that `Atom::Nil` exists (#51).
- The session's watermark and rewind track the new pool.
- `ElabCtx::add_const(Value)` is gone: nothing called it.

Int stays `i64` here, which is today's behaviour. Arbitrary-precision Int (decided in `docs/semantics.md`) is its own later change, and `Const::Int` is where it will land.

## Checks

- `cargo check -p scarlet_core`: down from 19 errors to 17. No `Value` is left in `typed_ir/` or `core_ir/`; the one remaining mention is the bytecode emitter's import, which goes in the next PR.


alii:
Squashed into `rewrite` as 53cd6c0 (`constants are front-end values`), one commit per PR. It lands on master with #42.

PR #54, opened 2026-09-20
drop the jit hook

alii:
Stacked on #53.

The JIT used to plug into the compiler through a callback, `NativeHook`. It fired once for each lowered function body, and was handed that body's core IR, its type pool, and a way to count the variants a `match` switches on. The Cranelift backend it fed was deleted in #42. This removes the hook itself.

## What

- `NativeHook` and `compile_with_native` are gone, and so is `CompileOptions::native_hook`.
- `Compiler` loses three fields:
  - `native_hook`
  - `native_stats`, the per-unit timing summary printed under `SCARLET_NATIVE_DEBUG`
  - `frame_layouts`: the slot layout the bytecode emitter chose for each body, kept so the JIT could agree with it
- `Emitted` loses `frame_layouts`.
- The two places the hook fired (`elaborate_body` and `materialize_eta_wrappers`) are gone. So is sizing `program.native`, the native-entry table.
- `core_ir::SwitchCounts` goes, since only the hook used it. `switch_variant_count` stays, because the bytecode emitter still calls it until the next PR.
- The `stdlib_native_gate_probe` test module goes. It asked how many stdlib bodies the JIT would accept.

A JIT for the new VM can come back later. The core IR plus Perceus is the natural thing to hand it, so the seam it needs is the one this PR removes, rebuilt against the new VM.

## Checks

- `cargo check -p scarlet_core`: down from 17 errors to 10, with none about the JIT


alii:
Squashed into `rewrite` as 0c3f032 (`drop the jit hook`), one commit per PR. It lands on master with #42.

PR #55, opened 2026-09-20
reuse shapes drop heap tags

alii:
Stacked on #54.

Perceus pairs a `Drop` with a later `Ctor` of the same shape, so the dying cell can be overwritten in place. The shape was `ReuseShape { tag: HeapTag, words: u16 }`, using the old VM's heap tag. But every shape was built by `ReuseShape::enum_(arity)`, so the tag was always `HeapTag::Enum` and never told two shapes apart.

Now it's just `ReuseShape { fields: u16 }`, a constructor cell with that many fields, and the constructor is renamed `enum_` → `ctor`. Pairing behaves exactly as before.

With this, `core_ir/` and `typed_ir/` no longer import anything from the bytecode layer or the old VM.

## Checks

- `cargo check -p scarlet_core`: down from 10 errors to 9. What's left is the bytecode emitter and the crate's leftover re-exports, which are the next PRs.


alii:
Squashed into `rewrite` as d9a4702 (`reuse shapes drop heap tags`), one commit per PR. It lands on master with #42.

PR #56, opened 2026-09-20
drop the abi templates

alii:
Stacked on #55.

At the end of a compile, the old compiler ran `bind_abi`. That baked a template for each stdlib constructor the old VM builds by itself (`Ok`, `Err`, `NetError` and so on) into `Program.templates`, and recorded where that prefix ended (`abi_template_count`), so a rewind could keep it and drop the rest. The code that built those templates (`bytecode/compiler/abi.rs`) went in #42. This removes what was left pointing at it:

- the `bind_abi` call at the end of `compile_with`
- the `Compiler::abi_template_count` field
- the session rewind's truncation of `program.templates`, `program.wire_templates` and `program.wire_descs`
- the two tests for that prefix (`abi_prefix`, and `reset_to_truncates_templates_to_the_abi_prefix`)

How the new VM will build stdlib values like `Ok(x)` is its own design question, for its own PR.

## Checks

- `cargo check -p scarlet_core`: still 9 errors, all in the bytecode emitter the next PR removes


alii:
Squashed into `rewrite` as 3cea94d (`drop the abi templates`), one commit per PR. It lands on master with #42.

PR #57, opened 2026-09-20
compiler stops at core ir

alii:
Stacked on #56. This is the big one: the compiler's output is now a complete program in core IR, not bytecode. It's about −960/+710 lines, most of that deletion. Here is what to read, in order.

## 1. The new output: `core_ir::Program` (`core_ir/mod.rs`)

This is the contract a new backend reads.

```rust
pub struct Program {
    pub fns: TiVec<FuncIdx, LoweredFn>, // every function: declared, lambdas, eta wrappers
    pub consts: Vec<Const>,
    pub inits: Vec<LoweredFn>,          // each imported module's top level, in run order
    pub toplevel: LoweredFn,            // the entry file's top level
    pub main: Option<FuncIdx>,          // `pub fn main`, when there is one
    pub globals: u32,                   // global slots module-scope bindings use
}
pub struct LoweredFn { pub name: String, pub core: CoreFn, pub pool: Rc<ResolvedPool> }
```

To run it: every `init` in order, then `toplevel`, then `main`.

Three things the old `CoreProgram` never had, because the emitter didn't need them:

- **Complete.** Imported modules' top levels used to go straight to bytecode, eta wrappers never reached core IR, and `fns` was in elaboration order. Now every function sits at its `FuncIdx`.
- **Typed.** Each body keeps the type pool its `RTy`s index. The old compiler threw the pool away after emitting, so a backend can now tell, for example, which locals are Ints.
- **Named.** `LoweredFn::name` is the real source name. `core.name` is an interned id that means nothing once the compile ends, and crash reports and stack traces need the real one.

## 2. One rule moved from the emitter into the IR: `perceus::keep_globals`

Perceus sees a module's top level on its own, so it drops a module-scope binding at its last read *in the top level*. But function bodies go on reading it through `Load::Global` for as long as the program runs. The old emitter quietly skipped those drops (`emit_drop`: "a pinned id emits nothing"). That rule is now a small core IR pass, run on every top level straight after Perceus. It removes the `Drop` of each pinned global, and any constructor `reuse` of its cell. So the IR is correct by itself, and a backend doesn't need to know the rule. It has a unit test.

## 3. What the compiler lost (`bytecode/compiler/mod.rs`)

- `program: Program` and `frozen: FrozenBuilder` are replaced by `fns: TiVec<FuncIdx, Option<LoweredFn>>`, `inits`, `entry_toplevel` and `main`. A function slot is reserved with `fns.push(None)` and filled once its body is lowered.
- `finish_program` assembles the `Program` only when the entry top level was recorded, which is still the `EntryToplevel` witness. An unfilled slot at that point is an internal-error abort (`function_never_lowered`).
- `append_toplevel_init` becomes `record_toplevel`, and `emit_module_init` becomes `lower_module_init`.
- Removed outright: all bytecode emission (`emit`, `current_addr`, `op*`, `Function` entries, `Halt`, `CallKnown` to main, `fuse`), and the jump-over machinery (`FnFrame::jump_over`, `DeferredBody::jump_over`, `region_jump_overs`, the drain's patch loop, `region_collector_lost`, `close_empty_deferred`). Also `param_slots`, `code_mark` and `slot_base`, and `CompileResult::into_artifacts`.
- `finish_fn_frame` loses its `name` and `arity` parameters, which existed only to fill a bytecode `Function` entry.

## 4. Around it

- `bridges.rs`: the `EmitCtx` impl goes, along with `declared_variant` and `switch_variant_count`.
- `session.rs`: the watermark drops `code`, and `functions` now tracks `Compiler::fns`. A rewind truncates `fns`/`consts` and drops recorded top levels. The two tests about the old `core` field are replaced by one about the new rule.
- Tests (`compiler/tests.rs`): the Perceus drop/reuse tests, "starts at `main`", "a script leaves its tail" and the wire-intrinsic test now assert on the lowered `Program` instead of opcodes. The three-argument `compile(…, None, None)` calls left over from #48 are fixed.
- The docs that described the pipeline now say `lower → perceus`. Comments elsewhere that still mention old opcode names (`PushGlobal` and so on) are wording only, and get a doc-only PR later.

## Checks

- `cargo check -p scarlet_core`: 9 errors left, none from this PR:
  - `restricted_generalization_cons` is missing. That's my mistake in #42: this type-checker function lived in `abi.rs`, which #42 deleted wholesale. The next PR restores it.
  - `Const::is_float` in `patterns.rs`, a leftover from #53.
  - `ReuseShape`'s old fields in the IR printer and one Perceus line, leftovers from #55.
  - The last `scarlet_vm` re-exports in `lib.rs` and `bytecode/mod.rs`.

  Those errors only became visible now: until this PR, name-resolution errors stopped rustc before it type-checked anything.


alii:
Squashed into `rewrite` as 6d978dc (`compiler stops at core ir`), one commit per PR. It lands on master with #42.

PR #58, opened 2026-09-20
fix leftovers the checker reaches

alii:
Stacked on #57.

Until #57, `scarlet_core` failed name resolution, so rustc never type-checked it. That hid three mistakes from earlier PRs in this stack. This fixes them.

## 1. The value restriction was deleted with `abi.rs` (my mistake in #42)

`restricted_generalization_cons` is part of the type checker. It is the relaxed value restriction: a `Subject(msg)`'s type parameter must not be generalized, or one mailbox could be sent one type and received as another. It happened to live in `bytecode/compiler/abi.rs`, which #42 deleted wholesale as VM ABI binding. It's restored unchanged into `compiler/mod.rs`, except that it looks the module up with `ModuleTable::get`: `get_or_hydrate` went in #48.

The audit found this restriction is incomplete: a Subject wrapped in a closure or a user type still escapes. That fix is its own later PR; this one only puts back what was there.

## 2. `Const::is_float` in `patterns.rs` (a leftover from #53)

The pattern checker still asked the old `Value` whether a literal is a float. It now uses `matches!(v, Const::Float(_))`, as the expression checker already does.

## 3. `ReuseShape`'s old fields (a leftover from #55)

- The core IR printer shows a drop's shape as `drop %0 [ctor:2]`. It used to print `[Enum:2]`, from the tag that no longer exists.
- Perceus's "a zero-field cell has nothing to reuse" check reads `fields` instead of `words`.

## Checks

- `cargo check -p scarlet_core`: down from 9 errors to 2. Those two are the `scarlet_vm` re-exports in `lib.rs` and `bytecode/mod.rs`, which the next PR removes.


alii:
Squashed into `rewrite` as 8f44558 (`fix leftovers the checker reaches`), one commit per PR. It lands on master with #42.

PR #59, opened 2026-09-20
drop the vm re-exports

alii:
Stacked on #58.

The last two references to the old VM in `scarlet_core`:

- `lib.rs` re-exported `scarlet_vm::{assert_send, assert_send_sync, frozen, heap}`. Nothing in the crate used them.
- `bytecode/mod.rs` re-exported `scarlet_vm::bytecode::*` (the ISA, `Program`, `Value`), so that "`scarlet_core::bytecode::*` is the one import for both halves". The compiler no longer produces bytecode, so there is only one half. The module doc now says what the module does, and that the name `bytecode` is historical.

`lib.rs` also goes from `#![deny(unsafe_code)]` to `#![forbid(unsafe_code)]`. The one allowed exception was in `core_ir::clif`'s tests, and #42 removed that file.

## Checks

- **`cargo check -p scarlet_core --all-targets`: compiles**, library and tests, for the first time on this stack. There are 13 warnings, all code that earlier PRs left unused (an import or two, and helpers only the emitter or the baked stdlib called). The next PR clears them so clippy's `-D warnings` can pass.


alii:
Squashed into `rewrite` as 55558da (`drop the vm re-exports`), one commit per PR. It lands on master with #42.

PR #60, opened 2026-09-20
the compiler crate is green

alii:
Stacked on #59. **The compiler crate is now green: it builds, its tests pass, and clippy and dylint are clean.** This is the first time on this stack.

## What

With #59 the crate compiled, but with 13 warnings (which clippy's `-D warnings` turns into errors) and 1 failing test. Everything here is code earlier PRs left unused:

- `ExportedValue::{param_names, doc}` and `ExportedType::{def, doc}`: only the hydrated-module definition synthesis read them, and that went in #49. The export helpers stop taking them, and `Prepared::doc` and two `def` locals go with them.
- `module::stdlib::all_modules`: only the stdlib pre-baker used it.
- `CtorRef::is` and the test-only `PreludeBindings::test_bool_binary`: their only callers were in the old emitter and its tests.
- `Compiler::diagnostics`: no callers left.
- Three unused imports.
- `Compiler::prelude_bindings`, `ResolvedPool::prims` and `CoreBind::id` are now only used by tests, so they're marked `#[cfg(test)]`.
- `let_drop_spine_is_flat` expected the printer's old `[Enum:3]`. It now expects `[ctor:3]`, the format #58 introduced.

## Checks

Scoped to the three crates that make up the front end: `scarlet_syntax`, `scarlet_types` and `scarlet_core`.

- `cargo fmt --all --check`: passes
- `cargo clippy -p scarlet_syntax -p scarlet_types -p scarlet_core --all-targets -- -D warnings`: passes
- `cargo test -p scarlet_syntax -p scarlet_types -p scarlet_core`: **467 passed**, 0 failed (compiler 206, syntax 191, types 70). That includes this stack's new and rewritten tests: `keep_globals`, the four Perceus drop/reuse tests (which now run through the real pipeline), starts-at-`main`, the script tail, wire as intrinsics, the session rewind, `binop`, and the intrinsic list.
- `cargo dylint --all -- -p scarlet_syntax -p scarlet_types -p scarlet_core` with `DYLINT_RUSTFLAGS=-D warnings`: passes
- Not yet: the driver crate `scarlet`, so workspace-wide checks, hawk and the golden tests. The driver still links the old VM and is the next piece of work.

PR #61, opened 2026-09-20
list a module's exports

alii:
Stacked on #60. **The compiler can now tell you what a module exports.** The REPL's tab completion used this list, and needs it back. Nothing uses it yet; the driver PR after this one will.

## Why

The REPL used to read stdlib names from the pre-baked stdlib blob. The blob is gone, since the stdlib compiles from source now (#48). So the list has to come from the compiled module itself.

## What

- `IncrementalSession::exports(path)` returns a module's exported names, types first, each with its parameter names. Example: `replace(s, pattern, replacement)`.
  - If no check has imported the module yet, it checks a one-line `import <path>` entry first. The module then lands in the cache through the normal import door, so the session's rewind bookkeeping stays true.
  - Constructors are listed with their field labels, e.g. `Fixed(len, b)`. A private constructor is not listed, because the interface already keeps those out.
  - A record's constructor has its type's name, so it is listed only once.
- `module::stdlib_modules()` lists every embedded stdlib module path, sorted. This is for completing `import scarlet/…`.
- Built-in (`@vm`) functions now remember their parameter names. `ValueKind::Builtin` gains `param_labels`, the same field `ModuleFn` already has. Without it, `println` would complete as `println` instead of `println(x)`.
  - I put the names on the scheme, not back on `ExportedValue`. #60 removed `ExportedValue::param_names`, and `ModuleFn`/`Constructor` already carry labels on the scheme. So the scheme stays the one place a function's parameter names live.

## Checks

Scoped to the three front-end crates, as in #60.

- `cargo test -p scarlet_syntax -p scarlet_types -p scarlet_core`: **472 passed**, 0 failed. That includes 5 new session tests: function and `@vm` params, prelude listed without a check, constructor labels and dedup, unknown module, and the module list.
- `cargo dylint --all -- -p scarlet_syntax -p scarlet_types -p scarlet_core` with `DYLINT_RUSTFLAGS=-D warnings`: passes
- Not yet: the driver crate, and so hawk. Hawk would flag the two new `pub` fns until the driver uses them, which is the next PR.


alii:
Squashed into `rewrite` as 17966b3 (`list a module's exports`), one commit per PR. It lands on master with #42.

PR #62, opened 2026-09-20
the driver builds again

alii:
Stacked on #61. **The `scarlet` binary builds again.** `check`, `fmt`, `lint`, the LSP and the REPL work. `run` and `dis` say plainly that the VM is being rebuilt.

## What

- **The pre-baked stdlib is gone from the driver.** `lib.rs` no longer includes `stdlib_generated.rs` or re-exports `STDLIB`. `IncrementalSession::new()` takes no stdlib argument.
- **`build.rs` comes back, smaller.** #42 deleted it whole, but it did two jobs. One was baking the stdlib, which is gone. The other was stamping `SCARLET_VERSION` for `scarlet --version`. This restores only the version stamping, unchanged.
- **REPL tab completion reads the compiled stdlib.** `Names` holds an `IncrementalSession` and asks it for exports (#61). The session is built on the first completion that needs it, and it caches each module it compiles, so each module compiles once per REPL.
  - Import aliases now map to a `ModulePath` instead of a key string.
  - Completion takes `&mut Names`, because asking can compile a module.
- **`scarlet run`** still compiles the file, so you see every error `check` would show. Then it exits 1 with `cannot run: the VM is being rebuilt`.
- **`scarlet dis`** also compiles, then exits 1 with `no bytecode to show: the VM is being rebuilt`. `--native` is gone, since the JIT is gone.
- **The REPL** compiles each entry. An entry that compiles joins the session, so its definitions stay usable. If the entry ends in an expression, it prints `not evaluated: the VM is being rebuilt`. `:type` works. `:dis` prints the same "no bytecode" line.
- Deleted: the native-code plumbing in `main.rs` (`publish_native`, `install_lazy_compiler`, `compile_one`) and the REPL's `is_unit`.

## Tried by hand

- `scarlet check` on a good file exits 0. On a type error it exits 1 with the error.
- `scarlet run` on a good file prints the rebuild message and exits 1. On a bad file it prints the type error first.
- `scarlet fmt --stdout`, `scarlet lint examples` and `scarlet --version` work.
- REPL: `const x = 1 + 2`, then `x`, then `:type x`. That prints `not evaluated…`, then `x  Int`.

## Checks

- `cargo clippy -p scarlet --lib --bins -- -D warnings`: passes. Also passes for the three front-end crates, as in #61.
- `cargo test -p scarlet --lib --bins`: **39 passed**, 0 failed. That includes all the completion tests: `println(x)` with no session, `http.Post` behind an alias, `import scarlet/str…`.
- `cargo dylint --all -- -p scarlet --lib --bins` with `DYLINT_RUSTFLAGS=-D warnings`: passes
- Not yet: the driver's integration tests (`crates/scarlet/tests/`). They still use the old VM's internals, so `cargo test --workspace`, `--all-targets` clippy and hawk can't build yet. I tried hawk: it builds every target, and it stops at those test files before it gets to analysis. The tests are the next PR.


alii:
Squashed into `rewrite` as c877783 (`the driver builds again`), one commit per PR. It lands on master with #42.

PR #63, opened 2026-09-20
drop the vm's own tests

alii:
Stacked on #62. **First of three test PRs.** This one deletes the driver tests that only make sense for the old VM's insides.

## Deleted

Each of these tests a piece that #42 removed. None of them tests what a Scarlet program means.

- `dis.rs`: the bytecode listing and its jump targets. There is no bytecode now.
- `jit_spike.rs`, `native_hook.rs`, `native_freed_balance.rs`: the Cranelift JIT, its compile hook, and its heap-accounting parity with the interpreter.
- `vm_fairness.rs`: the reduction budget (4000 per slice) in the old scheduler.
- `vm_exec.rs`, second half: the tests that ran the VM in-process to read `ProcHeap::alloc_count()`, or read opcodes off the bytecode. Examples: list `map` reusing cells, the `dot_loop` reuse gate, `a[0] or 0` allocating nothing.

All of these are on `master` if the new VM wants to borrow one back.

## Kept

- `vm_exec.rs`'s black-box tests stay. These run a program through `scarlet run` and compare what it prints. That's the closures, `and`/`or`, equality, both `index_or` behaviour tests, `bench_typed`'s pinned output, and the `run_case!` blocks. They say what programs mean, so the new VM has to pass them too. For now they fail at run time; the third test PR parks them.
- **Moved, not deleted:** two `vm_exec` tests checked that `+`, `-` and `==` on `Int` pick the typed op (`AddInt`), never the generic one (`Add`). That is a compiler fact, not a VM fact. So they now live in the compiler's own tests (`typed_prim_ops` in `compiler/tests.rs`) and read the same fact off Core IR: `PrimOp::IntAdd` present, `PrimOp::Add` absent. One of them pins the tricky case, where only inference knows the operand is an `Int`.
  - I checked they can fail. Making int `+` lower to `PrimOp::Add` turns both red.

## Checks

- `cargo test -p scarlet_syntax -p scarlet_types -p scarlet_core`: **474 passed**, 0 failed. That's 472 from #61 plus the 2 moved tests.
- `vm_exec.rs` builds again.
- Not yet: most other driver test files still don't build. Almost all of them trip on one line in the shared helper `common/mod.rs`, and 8 also call the old compile API in their own code. The next PR fixes those.


alii:
Squashed into `rewrite` as cecd2e5 (`drop the vm's own tests`), one commit per PR. It lands on master with #42.

PR #64, opened 2026-09-20
functions know their module

alii:
Stacked on #63. **Every function in a compiled `Program` now says which module it came from.**

## Why

The stdlib now compiles from source into the same `Program` as your code (#48). So a function's name alone is ambiguous: `scarlet/array` and `scarlet/option` both have a `map`, and so can your file. Before, the compiler kept a separate copy of just the entry file's IR, and tools read that. That copy went in #57.

Two things need the answer now:

- The Core IR snapshot test (next PR) wants only your file's functions, not the whole stdlib's.
- Anything a person reads later, like a stack trace or `dis`, wants `scarlet/array.map` rather than `map`.

## What

- `LoweredFn` gains `module: ModuleKey`. That's the module's canonical identity, not the name you wrote in an import.
- One helper, `Compiler::lowered_fn`, builds every `LoweredFn`: function bodies, eta wrappers (small functions the compiler writes when you pass a constructor or built-in as a value), and toplevels. It reads the module being compiled at that moment. Bodies, wrappers and toplevels are only ever lowered while their own module is the one being compiled, so that's their module.
- The three places that used to build a `LoweredFn` by hand now call the helper, so none can forget the module.

## Checks

- `cargo clippy -p scarlet --lib --bins -- -D warnings`: passes
- `cargo test -p scarlet_syntax -p scarlet_types -p scarlet_core`: **475 passed**, 0 failed. The new test is `two_maps_are_told_apart_by_their_module`: a file defines its own `map` and also calls `array.map`, and the program holds one `map` from `main` and one from `scarlet/array`.
- `cargo test -p scarlet --lib --bins`: 39 passed
- `cargo dylint --all -- -p scarlet_syntax -p scarlet_types -p scarlet_core -p scarlet --lib --bins` with `DYLINT_RUSTFLAGS=-D warnings`: passes
- Not yet: the driver's integration tests. They're the next two PRs.


alii:
Squashed into `rewrite` as a7a800f (`functions know their module`), one commit per PR. It lands on master with #42.

PR #65, opened 2026-09-20
tests follow the new api

alii:
Stacked on #64. **Second of three test PRs.** Every driver test file builds again. The fixes here are to the tests only; no compiler code changes.

## What

- **Old calls updated.** `IncrementalSession::new(&STDLIB)` becomes `IncrementalSession::new()`, and `compile`/`check` drop the stdlib argument. That's the shared helper in `common/mod.rs` plus a handful of files.
- **Core IR snapshots (`core_ir.rs`)** print the `Program` now. They show only the entry file's functions, using #64's `module` field, and then the toplevel. All 15 `.core` snapshots are regenerated. I read the diff line by line:
  - Renames only: `AddInt` → `IntAdd`, `GetFieldUnchecked#0` → `FieldUnchecked(0)`, `BinBitSize` → `BinaryBitSize`, `[Enum:2]` → `[ctor:2]`, `PushNil()` → `nil`, `Str("PUT")` → `String("PUT")`. Binary constants print their bytes now. The entry module's type ids moved from `10241.x` to `512.x`, because the stdlib now claims its type-id ranges from source.
  - Real change 1: the toplevel no longer drops module globals (`drop %0` after `closure fn#0()`). That is #57's `keep_globals`, visible for the first time.
  - Real change 2: a binary pattern's offset maths is now typed. `Add`/`Eq` on the bit offsets became `IntAdd`/`IntEq`.
  - `the_same_program_without_the_error_node_compiles_and_runs` ran the VM. It now checks that the program is runnable and starts at `main`. The run part comes back with the VM.
- **`arena_rewind.rs`**: `stdlib_bodies_are_never_relowered` is deleted. It checked that the pre-baked stdlib's bodies were never lowered at run time. Now the stdlib compiles from source, so its bodies are lowered like any other. The file's doc comments no longer describe the blob.
- **Deleted `check_parity.rs`.** Half of it compared the function tables `check` and `compile` build. `check` no longer builds one: it stops before any function is recorded. The other half checked bytecode jump layout.
- **Deleted `stdlib_ctor_visibility.rs`.** It checked that constructor visibility survived the pre-baked stdlib blob. There is no blob. The same bit is checked where it's computed, in `ctor_visibility_survives_on_the_type` in the compiler's own tests.

## Checks

These are the CI commands, spelled the way CI spells them. This is the first PR on the stack where all of them run.

- `cargo clippy --workspace --all-targets -- -D warnings`: **passes**
- `cargo dylint --all` with `DYLINT_RUSTFLAGS=-D warnings`: **passes**
- `cargo test --workspace`: **fails**. The front-end crates pass all 475. The driver passes 384 and fails 331:
  - **300** fail only because nothing can run a program yet (`cannot run: the VM is being rebuilt`). The next PR parks those.
  - **31** are something else, and several are real bugs in this stack that the tests just caught:
    - `import scarlet/map` then a bare `Map(String, Int)` says `Unknown type 'Map'`. `map.Map` works. About 10 tests.
    - Two rewind tests crash: `range end index 16 out of range for slice of length 14` in `InferEngine::str_ids_of`. An index outlives the arena it points into.
    - `wire.decode` with nothing to fix its type is no longer refused by `check`. 3 tests.
    - `incremental.rs` counts 3 module compiles where it expects 2. My guess is the stdlib, which now compiles as a real module. I haven't checked yet.
    - REPL tests that expect a printed value, one `dis --native` test, and 4 supervision tests whose output check fails before it can show the VM message.
  - Each real bug gets its own PR after the next one.
- `cargo hawk check -D warnings`: **fails, 57 findings**. It runs for the first time. Most are fields that were `pub` only so the old stdlib baker or the old VM could read them, like `InferEngine`'s arenas. They get their own PR.
- grammar (`bun`): not run. Nothing here touches `editors/`.


alii:
Squashed into `rewrite` as 1abbe7c (`tests follow the new api`), one commit per PR. It lands on master with #42.

PR #66, opened 2026-09-20
run tests wait for vm

alii:
Stacked on #65. **Third of three test PRs.** Tests that need to run a program are now ignored with the reason `needs the VM`, so `cargo test` shows only real failures.

## Why ignore, not delete

These tests say what Scarlet programs mean: what they print, how closures capture, what `a[i] or x` does at the edges. The new VM has to pass them. Ignoring them keeps them in the tree as its to-do list:

```
cargo test -p scarlet -- --ignored
```

Each VM PR can then remove the `#[ignore]` from exactly the tests it makes pass, and the diff says so. Some will need changing when they come back: the crash tests in `vm_exits.rs` describe crashing, which the new design rules out.

## What

- **191 test functions** get `#[ignore = "needs the VM"]`. I picked each one from the test log: it failed with `cannot run: the VM is being rebuilt`, or it's one of the 13 below that I checked by hand.
- **`run_case!`** (in `common/mod.rs`) makes a run test from every case, so the ignore goes on the macro. That's 78 cases.
- **Golden examples:** the 45 run-and-diff tests are ignored in the `suite!` macro. A new test, `every_golden_program_checks`, runs `scarlet check` over every one of those programs. So a type-checker bug can't hide behind the ignore. It's already doing its job: it fails on `interpreter.scrl` because of the `Map` bug (#65).
- **Checked by hand:** 8 REPL tests that expect a printed value, 4 supervision tests, and one wire closure test. None of their failure messages include the program's stderr, so the log alone didn't show the VM message.
- `dis_lists_only_the_matching_function` (REPL `:dis`) is ignored with its own reason, `waits for :dis to print Core IR`. That comes back when `dis` prints Core IR.
- Deleted `dis_native_prints_clif_for_fib`. `--native` and the JIT are gone.

## Checks

- `cargo dylint --all` with `DYLINT_RUSTFLAGS=-D warnings`: passes (on #65; this PR only touches tests)
- `cargo test --workspace`: front-end crates 475 passed. Driver: **384 passed, 16 failed, 315 ignored**. The 16 are exactly the real bugs #65 lists, and each gets its own PR next:
  - `Map` unknown without `map.`: `every_golden_program_checks`, `field_access_through_a_module_fn_inferred_scrutinee`. Fixed in #70.
  - arena index past the end after a rewind: `stdlib_prefix_survives_repeated_rewinds`, `incremental_edit_b_keeps_refs_then_invalidate_drops_reverse_edges`. Fixed in #67.
  - module compile counts: 6 tests in `incremental.rs`. Fixed in #68.
  - `wire`'s type refusals, 6 tests (3 in `type_errors.rs`, 3 in `wire_descriptor.rs`). Not a new bug: #50 paused them on purpose. Parked in #69.
- `cargo hawk check -D warnings`: still 57 findings, as in #65
- grammar (`bun`): not run. Nothing here touches `editors/`.


alii:
Squashed into `rewrite` as 0d21cb0 (`run tests wait for vm`), one commit per PR. It lands on master with #42.

PR #67, opened 2026-09-20
rewinds evict later stdlib modules

alii:
Stacked on #66. **Fixes a crash in the LSP session:** after you edit a module, a stdlib module it didn't depend on could be left pointing at memory the rewind had just thrown away.

## The bug

`IncrementalSession` (the LSP's compiler) keeps compiled modules cached. When you edit `lib.scrl`, it rewinds its arenas to the point just before `lib` was compiled, then evicts `lib` and everything that imports `lib`.

That's only safe if nothing *else* above that point stays cached. The eviction kept every stdlib module unconditionally, because only `File` modules recorded a watermark (where their arena state starts). That held while the stdlib was pre-baked below everything. Since #48 it compiles from source like any module, in import order. So:

```
import ./lib             // compiled 1st
import scarlet/array     // compiled 2nd: above lib's watermark
```

Editing `lib` rewound past `scarlet/array`, kept its cached interface anyway, and the next lookup of `array.map` read a `DefinitionLocation` slice past the end of the arena:

```
range end index 16 out of range for slice of length 14   (InferEngine::str_ids_of)
```

## The fix

- The watermark moves from `ModuleOrigin::File` onto `CachedModule`, so every cached module has one. An embedded module's *source* never changes, but its interface still indexes the arenas it was compiled into.
- `invalidate(key)` evicts `key`, everything that depends on it, **and every module compiled after it**. That's the modules whose watermark isn't below the rewind point.
- `invalidate_all()` (the fallback when a module's type-id range overflows) keeps only the modules compiled before the first user module. Before, it kept every stdlib module, with the same bug.
- An evicted stdlib module is simply compiled again the next time something imports it.

## Checks

- `cargo test --workspace`: **14 failed**, down from 16 on #66, 315 ignored. The two that crashed now pass: `stdlib_prefix_survives_repeated_rewinds` and `incremental_edit_b_keeps_refs_then_invalidate_drops_reverse_edges`. The other 14 are the known bugs from #66, fixed in the next PRs.
- Two new unit tests in `module/mod.rs`:
  - `invalidating_a_module_evicts_every_module_compiled_after_it`: `string` (before `lib`) stays; `lib` and `array` (after) go. I checked it fails if embedded modules are kept again.
  - `invalidating_everything_keeps_only_what_came_before_the_first_file`
- `cargo hawk check -D warnings`: not rerun. It's still expected to show the 57 findings from #65; this PR adds no `pub` items.


alii:
Squashed into `rewrite` as 955f112 (`rewinds evict later stdlib modules`), one commit per PR. It lands on master with #42.

PR #68, opened 2026-09-20
count only what checks compile

alii:
Stacked on #67. **`IncrementalSession::compile_count()` counts only what checks compile.**

## Why

`compile_count` is telemetry: how many module bodies were actually compiled, not cache hits. The incremental tests use it to prove that editing `c.scrl` recompiles `c` and `b` but not `a`'s other imports.

Since #48 a new session compiles the prelude from source when it starts. So every count started at 1, and 6 tests in `incremental.rs` saw `3` where they meant "b + c compile, that's 2".

The tests are right about what matters. A session's count answers "what did my checks compile", and the prelude is compiled by `new()`, not by a check. So the session records how many bodies seeding compiled and leaves them out. The module table's own counter still counts everything.

## Checks

- `cargo test --workspace`: **8 failed**, down from 14 on #67. All 9 tests in `incremental.rs` pass now. New unit test: `a_new_session_counts_no_compiles`. The 8 left are the `Map` and `wire.decode` bugs.
- `cargo hawk check -D warnings`: not rerun; this PR adds no `pub` items.


alii:
Squashed into `rewrite` as 2221188 (`count only what checks compile`), one commit per PR. It lands on master with #42.

PR #69, opened 2026-09-20
park the wire refusals

alii:
Stacked on #68. **Parks the 6 tests for `wire`'s compile-time refusals, with the reason on each.**

## Why these fail, and why that's expected

`wire.decode` needs to know what type it's decoding, and `wire.encode` can't send some types (a closure, say). `scarlet check` used to refuse those programs with messages like:

> the type `wire.decode` produces here is not known; annotate the binding

That check lived in the old VM's *descriptor builder*: a description of the type, attached to the VM instruction. #50 removed the descriptors along with the instructions, and said so under "`wire` loses its descriptors, for now": the check comes back with the new VM's wire design, which also has the audit's wire findings to fix (forged opaque types, tampered closures).

So these 6 tests describe a feature that's paused on purpose. They're not a new bug.

## What

`#[ignore = "waits for the wire redesign: check stopped refusing wire types in #50"]` on:

- `type_errors.rs`: `unknown_payload::an_unconstrained_decode_is_refused`, `unknown_payload::a_decode_in_a_generic_fn_is_refused`, `wire_names_the_whole_path_down_to_the_refusing_type`
- `wire_descriptor.rs`: `a_refusal_is_reported_on_the_session_check_path`, `a_session_re_checks_a_wire_call_across_an_edit`, `the_refusal_path_survives_to_the_session_check_path`

Two of them also write a bare `Map(...)`, which the next PR deals with.

## Checks

- `cargo test`: `type_errors.rs` and `wire_descriptor.rs` pass, with those 6 ignored. The workspace is down to 2 failures, both the `Map` bug, fixed next.
- dylint and hawk: not rerun; this PR only touches test attributes.


alii:
Squashed into `rewrite` as 128530e (`park the wire refusals`), one commit per PR. It lands on master with #42.

PR #70, opened 2026-09-20
stdlib types need their module

alii:
Stacked on #69. **A stdlib type is only in scope through its import.** `Map` after `import scarlet/map` is `map.Map`, or `Map` with `import scarlet/map.{Map}`. And the error now tells you which spelling works.

## What was happening

On `master`, both of these type-check, with **no import at all**:

```
fn f(_m Map(String, Int)) Int { 0 }
fn g(_r Request) Int { 0 }
```

The pre-baked stdlib loaded every stdlib module's type names into one global table at startup, so all 85 stdlib types were in every program's namespace. That was a leak, not a rule: the import code already says to bind an imported type "through this module's env rather than residual global `type_info` left over from the dependency's own compile".

Since #48 the stdlib compiles like any module, and a module's names stay inside the module. So the leak closed on its own, and anything that leaned on it broke: `Unknown type 'Map'`.

## What leaned on it

I measured this rather than guessing: `scarlet check` on every `.scrl` in the repo, plus every parked test run with `--ignored` (`scarlet run` type-checks before it stops). **Only `Map`** ever relied on the leak. Every other stdlib type was already imported by name.

- `examples/interpreter.scrl`, `bench_map.scrl`, `bench_service.scrl`: `Map(...)` becomes `map.Map(...)`. `interpreter.scrl` is re-formatted where a signature got too long. `bench_service.scrl` was already unformatted before this and I left the rest of it alone.
- 4 test snippets: `map.Map`. Two of them never imported `scarlet/map` at all, so they gain the import.

## The error says what to write

```
error: Unknown type 'Map'. Did you mean 'map.Map'?
```

An import already registers each of its module's types under `qualifier.Name`. So when a bare name misses, the new `TypeEnv::qualified_spellings` looks up every `x.Map` in scope. It uses the real qualifier, so `import scarlet/map as m` suggests `'m.Map'`. With nothing imported that exports a `Map`, the message is the plain `Unknown type 'Map'` as before.

## Checks

- `cargo test --workspace`: **875 passed, 0 failed, 322 ignored.** First fully green test run on the stack. New tests:
  - `a_bare_type_from_an_imported_module_points_at_its_qualified_name`, which covers both the plain and the aliased import
  - `a_stdlib_type_is_unknown_without_its_import`, which pins the leak shut
  - `every_golden_program_checks` passes again
  - `field_access_through_a_module_fn_inferred_scrutinee` now type-checks. It also runs the program, so it joins the `needs the VM` tests.
- grammar, in `editors/tree-sitter-scarlet`: `bun install`, `bun run generate && git diff --exit-code src grammar.js`, `bun run corpus`: all pass
- `cargo hawk check -D warnings`: 57 findings, the same ones as #65. None are new.

## Follow-up

9 parked tests do a type-check *and then* a run, so parking them in #66 also switched off their type-check half. The next PR splits those so the check half runs again.


alii:
Squashed into `rewrite` as fb58d20 (`stdlib types need their module`), one commit per PR. It lands on master with #42.

PR #71, opened 2026-09-20
parked tests keep their checks

alii:
Stacked on #70. **9 parked tests had a type-check half that parking switched off. This splits them so the check half runs again.**

## Why

#66 parked every test that runs a program. A few of those also asserted what `scarlet check` does before running, for example `check_rejects(... "not exhaustive")` next to a `run_outputs`. `#[ignore]` covers the whole function, so those checks stopped running, even though they need no VM.

## What

Each of the 9 is split in two:

- `<name>` keeps its `check_ok`/`check_rejects` calls and runs again.
- `<name>_runs` gets the `run_outputs` calls and stays `#[ignore = "needs the VM"]`.

Where both halves used the same program (`let src = ...`), the source becomes a shared `const <NAME>_SRC`, so the two can't drift apart.

The 9, with how many calls went each way (check, run):

| test | check | run |
|---|---|---|
| `unsound::u21_exhaustiveness_respects_field_labels` | 1 | 1 |
| `stdlib::stdlib_binary` | 1 | 7 |
| `stdlib::stdlib_binary_slice_units` | 1 | 5 |
| `type_semantics::bool_is_a_normal_two_ctor_type` | 2 | 2 |
| `type_semantics::reserved_set_derived_from_prelude_iface` | 1 | 1 |
| `type_semantics::binary_string_literal_patterns` | 1 | 9 |
| `type_semantics::field_access_through_a_constructor_inferred_scrutinee` | 1 | 1 |
| `type_semantics::field_access_through_a_module_fn_inferred_scrutinee` | 1 | 1 |
| `type_semantics::or_receiver_binds_a_heap_error_payload` | 1 | 1 |

I found them by searching every `needs the VM` test body for a `check_ok(` or `check_rejects(` call, and moved the calls with a script, so no call was dropped or duplicated.

## Checks

- `cargo test --workspace`: **884 passed**, up from 875 (the 9 check halves), 0 failed, 322 ignored (the same count: the 9 `_runs` replace the 9 parked originals)
- dylint, hawk and grammar: not rerun; this PR only moves test code around.


alii:
Squashed into `rewrite` as 0c8a8c5 (`parked tests keep their checks`), one commit per PR. It lands on master with #42.

PR #72, opened 2026-09-20
dis prints core ir

alii:
Stacked on #71. **`scarlet dis` and the REPL's `:dis` work again. They print Core IR**, the compiler's output since #57, where they used to print old-VM bytecode.

## What it looks like

```
$ scarlet dis sq.scrl
; 2 functions, 2 from this file; 1 module inits; 2 globals; starts at fn#1 (main)

; fn#0
fn main.square(%0:6) -> :6
  ret IntMul(%0, %0)

; fn#1
fn main.main() -> :6
  let %0:0 = c0
  let %1:10 = call fn#0(%0)
  ret Println(%1)

; toplevel
fn main.__main__() -> :5
  let %0:8@g0 = closure fn#0()
  let %1:10@g1 = closure fn#1()
  ret nil
```

- `scarlet dis FILE` shows only this file's functions, then its toplevel. The stdlib compiles into the same program, and listing all of it by default would bury your code.
- `scarlet dis FILE --fn NAME` shows every function whose name contains `NAME`, from any module. For example, `--fn replace` finds `fn scarlet/string.replace(...)`. A name nothing has exits 1 with `no function matching 'NAME'`.
- `:dis NAME` in the REPL is the `--fn` form over the session's source.
- Each function is headed with its `fn#N`, the number `call fn#N` and `closure fn#N` refer to. Its name is `module.name`, using #64's `module` field, because two modules can both have a `map`.
- The first line summarises the whole program, so a filtered listing still says what it came from.

## How

- `core_ir`: `impl Display for LoweredFn` prints the same thing as `CoreFn`'s, but headed with `module.name` instead of the interned id (`fn s1234(`). Both go through one `write_fn` helper.
- `crates/scarlet/src/dis.rs`, new: `listing(program, Filter::Entry | Filter::Named(name)) -> Option<String>`. `None` means the filter matched nothing.
- `main.rs` and `repl/mod.rs` call it where they used to print "no bytecode to show".

## Two things I noticed, not fixed here

- **The `:6` type numbers are per function.** Each lowered body has its own type pool, so `:6` in `square` and `:6` in `main` can be different types. The golden snapshots hide this by renumbering. Printing real types (`%0:Int`) means passing the pool into the printer, which is its own PR.
- **Constants print as `c0`**, with no value next to them. The snapshots add a `where c0 = Int(3)` block; `dis` could too.

## Checks

- `cargo test --workspace`: **894 passed**, 0 failed, 321 ignored. New tests:
  - 4 unit tests on the listing: entry-only, reaching into the stdlib by name, no match, the summary line
  - 3 end-to-end tests in a new `tests/dis.rs`: lists, fails on an unknown name, reports a compile error instead of listing
  - the REPL's `dis_lists_only_the_matching_function`, parked in #66, runs again
- `cargo hawk check -D warnings`: **54 findings**, down from 57. `LoweredFn::name`, `Program::inits` and `Program::globals` now have a real reader.


alii:
Squashed into `rewrite` as 18fe00f (`dis prints core ir`), one commit per PR. It lands on master with #42.

PR #73, opened 2026-09-20
hawk is clean

alii:
Stacked on #72. **`cargo hawk check -D warnings` goes from 54 findings to 0.** With this, every CI check passes on the stack.

## What

Most of these were `pub` only so the old stdlib baker (`crates/scarlet/build.rs`) or the old VM could read them: `InferEngine`'s arenas, `EnginePoolWatermark`'s fields, `QuantVar`, `ArenaSlice`, `TypeInfo`, `DefinitionLocation`, `TiVec`'s methods, `PreludeBindings`' refs. Both readers are gone.

- **Narrowed:** about 50 items go to `pub(crate)`, then 26 of those to private, the tightest each one's real uses allow. I applied hawk's own findings, at the exact line and column it reported, and ran it again until it found nothing. Nothing was narrowed by guesswork.
- **Deleted, dead with no caller anywhere:** `ModuleKey::prelude`, `TypeEnv::declaration`, `TiVec::{get, get_mut}`.
- **Deleted, written but never read:** once they were `pub(crate)`, rustc could see these are only ever written, never read.
  - `CtorRef::arity`. `capture` still checks each prelude constructor's arity; it just stopped storing the answer nobody read.
  - `TiVec::{contains_idx, resize_at_least}`, and the test of them.
  - `TypedProgram::consts` and `CoreProgram::consts`. The compiler always passed an **empty** list here, and `lower` copied it into `CoreProgram` for the old emitter to "adopt back". Every `ConstId` indexes the compiler's one pool, `Program::consts`. The two tests that pinned the copy now pin what's left: `lower` keeps each constant's id (`ret c1`, `ret c0`) and never makes a new one (no `c3` when three are pooled).

## One field kept on purpose

`LoweredFn::pool` has no reader yet. But every type in a function's Core IR (`%0:6`) is an index into that pool, so dropping it would leave the `Program` full of indices into an arena that no longer exists. It stays `pub(crate)` with:

```rust
#[expect(dead_code, reason = "the RTys in `core` index it; the backend that reads them is not written yet")]
```

`#[expect]`, not `#[allow]`: the day a backend reads the field, the lint stops firing, `expect` turns that into an error, and the note can't go stale.

## Checks

All seven CI steps:

- `cargo test --workspace`: **893 passed**, 0 failed, 321 ignored. That's one fewer than #72: the deleted `resize_at_least` test.
- `cargo hawk check -D warnings --target-dir target/hawk`: **`hawk: 0 finding(s)`**. This is the host target, `aarch64-apple-darwin`. CI analyses Linux. None of the items changed here sits behind a `cfg(target_os)` or `cfg(unix)`, so the two should agree, but I haven't run the Linux cross-analysis.
- grammar (`bun install`, `generate` + diff, `corpus`): passed on #70. Nothing since touches `editors/` or a `.scrl`.


alii:
Squashed into `rewrite` as 8a11ede (`hawk is clean`), one commit per PR. It lands on master with #42.

PR #74, opened 2026-09-20
write down the vm plan

alii:
Stacked on #73. **Adds `docs/vm-design.md`: the plan for the new VM, written before any VM code.** No code changes.

It's marked like `docs/semantics.md`: **Decided** (you chose it), **Proposed** (my recommendation, waiting for your yes), and **Open**.

## What's in it

- **What the VM is given:** `core_ir::Program`, field by field, and the handful of Core IR pieces it has to run.
- **The rule the VM keeps:** a table of every way the old VM let code crash a process or the whole program, and what the new one does instead. Plus one proposed rule that closes the whole class: the VM never recurses in Rust on anything the program controls.
- **Values:** one 64-bit word; floats stored as themselves; big ints on the heap, with each number stored exactly one way; the no-NaN float rule, including why `x / 0.0` needs its own check before dividing.
- **Memory:** reference counting and Perceus, and how they split the work. The VM counts references itself, so memory is correct even where Perceus inserts nothing, and Perceus makes releases earlier and allows in-place reuse. This is how the old VM worked, and it answers a real question: Perceus never drops `Int`, `String`, `Array` or generic values, by design (`ResolvedPool::is_heap`), so a big `Int` would leak if the VM relied on Perceus alone.
- **Running code, where the VM lives, processes,** and **the order of work**. Each VM PR removes the `#[ignore]` from the tests it makes pass, so `cargo test -p scarlet -- --ignored` (321 today) is the progress meter.

## Three decisions, now made

1. **Per-process memory: a heap per process.** The doc now says how that works with reference counting. The heap is plain data the process owns: chunks with a free list per cell size, so it moves with the process between threads. A few things live outside any one heap: constants, shared big binaries, OS handles. A dead process releases those first ("off-heap list", as in BEAM), then returns its chunks whole.
2. **Execution: flat register instructions per function, made at load.** A JIT comes later and compiles the same instructions, so the interpreter and JIT agree by construction.
3. **Crates: a new `scarlet_ir` crate** holds the Program types, and `scarlet_vm` depends only on it.

One new open question came with 1: how big a process heap starts and how it grows, since a million tiny processes with large first chunks would waste memory.

## Checks

Docs only. Nothing to build or test.


alii:
Squashed into `rewrite` as a42c0bd (`write down the vm plan`), one commit per PR. It lands on master with #42.

PR #75, opened 2026-09-20
start the ir crate

alii:
Stacked on #74. **Step 1 of the VM plan: the `scarlet_ir` crate exists.** It's the contract between the compiler and the VM: the vocabulary a compiled program is written in. It depends on nothing in the workspace, and the VM will depend only on it. That's the "where the VM lives" decision in `docs/vm-design.md`.

This PR moves the three smallest pieces, the ones everything else is built from. No behaviour changes.

## What moves

- **`TypeId`**, from `scarlet_types::type_def`: a type's identity. A constructor is a `TypeId` plus a variant number, so the VM needs it as much as the type checker does.
- **`Intrinsic`**, from `scarlet_types::intrinsic` (`git mv`, unchanged): the 135 built-ins.
- **`TiVec`, `Idx` and `newtype_index!`**, from `scarlet_core::tivec` (`git mv`): a vector that only its own index type can subscript. `Program::fns` is one.

## What doesn't change

- Every old path still works, through re-exports: `scarlet_types::type_def::TypeId`, `scarlet_types::intrinsic::Intrinsic`, `scarlet_core::tivec`. So no code outside the crate roots had to change, apart from one `use` of the macro.
- `TiVec`'s methods go back from `pub(crate)` to `pub`, because the compiler now calls them from another crate. `next_idx` stays private: only `TiVec` itself uses it.

## The repo's instructions file

- Its crate-layout paragraph still described the old VM: bytecode, a JIT, a "language-agnostic" runtime, `scarlet_types` over `scarlet_vm`. It now describes the stack as it is: `scarlet_ir` and what it's for, the compiler stopping at Core IR, and no VM yet.
- "the same 43 test executables" is now 42. I counted both `cargo test --no-run` and `cargo test --workspace --no-run`.

## Checks

- `cargo test --workspace`: 893 passed, 0 failed, 321 ignored (unchanged)

## Next

The Program types themselves move next: the type pool (`RTy`, `ResolvedPool`), then the Core IR, `PrimOp`, `Const` and `Program`. One question comes with them: some Core IR fields hold `StrId`s, ids into the compiler's string table, which mean nothing once the compile is over. The VM shouldn't see those. I'll look at each one when it moves.


alii:
Squashed into `rewrite` as bcd0acf (`start the ir crate`), one commit per PR. It lands on master with #42.

PR #76, opened 2026-09-20
the type pool moves to ir

alii:
Stacked on #75. **The type pool moves into `scarlet_ir`.** Every type in a compiled function (`%0:6`) is a number into its pool, so the VM needs the pool as much as the compiler does.

## What moves

`scarlet_core/src/typed_ir/rty.rs` → `scarlet_ir/src/rty.rs`:

- `RTy`, `RSlice`, `ResolvedNode` and `ResolvedPool`, with their tests.
- `PrimIds` and `Prim`, from `scarlet_types`. The pool needs them to answer "is this an `Int`?". `scarlet_types` re-exports both under their old names (`PrimIds`, `PrimitiveKind`), so the type checker didn't change.

`Arity` stays in the compiler: nothing in the Program uses it. It gets its own small test, since the moved tests no longer check it.

## One real change: type nodes lose their name

`ResolvedNode::Con` carried `name: StrId`, an id into the compiler's string table. Its doc said "display only", but nothing displayed it. The one place that touched it (`subst_rty`) only copied it into a new node. It was dead weight, and it would have handed the VM an id into a table the VM never sees.

So `Con { id, args }` now, and `ResolvedPool::mk_con(id, args)` loses its name argument, at 31 call sites. The type checker's own `InferEngine::mk_con` is a different function and keeps its name. It does display it, in error messages.

## Smaller things

- The pool's methods are `pub` now, since the compiler calls them from another crate. `len` became `node_count`: clippy expects a public `len` to come with an `is_empty`, and `node_count` says what it counts anyway.
- `docs/vm-design.md` had one thing wrong: it listed `Array` among the types Perceus leaves alone. `PrimIds::prim_of` doesn't map `Array`, so arrays *are* dropped by Perceus. Only `Int`, `Float`, `String` and type variables are left to the VM.

## Checks

- `cargo test --workspace`: **894 passed**, 0 failed, 321 ignored. That's one more than #75, the `Arity` test.
  - On one run, two `vm_io` network tests failed: `connect_within_times_out_against_a_full_accept_queue` and `connect_addr_within_…`. Their setup depends on the host: it skips the test if the kernel resets a connection, and panics if it accepts all of them. They never run a program on this stack either way, so they belong with the parked tests. The next PR parks them.


alii:
Squashed into `rewrite` as 90a4fd3 (`type pool moves to ir`), one commit per PR. It lands on master with #42.

PR #77, opened 2026-09-20
park two network tests

alii:
Stacked on #76. **Parks three network tests that only ever passed by skipping themselves.**

`connect_within_times_out_against_a_full_accept_queue` and `connect_addr_within_times_out_against_a_full_accept_queue` first fill a `listen(1)` socket's accept queue, so that a later connect hangs, then run a Scarlet program against it. How the fill goes depends on the host's kernel:

- If the kernel resets the extra connections, the helper prints "unmeasurable" and the test returns early. It passes without running anything.
- If the kernel accepts all 16, the helper panics. That happened on one run of #76.
- Only if the kernel drops the SYN does the test go on to run a program. On this stack, that then fails with `cannot run: the VM is being rebuilt`.

#66 missed them because on that run they took the first path. They need the VM to measure anything, so they get `#[ignore = "needs the VM"]` like the others.

**Added later: a third one, in `vm_tls`.** `tls_connect_within_times_out_against_a_peer_that_never_accepts` uses the same accept-queue helper. I first wrote here that no other live test runs a program, and that was wrong: I had only looked on a Mac, where this test skips itself. On Linux CI it runs, and it failed on this PR and on every PR above it. It's parked now too, and the fix is merged up the stack.

Those three are the only tests that use the accept-queue helper, and no test has a Linux-only `cfg`, so nothing else runs differently on the two hosts.

## Checks

- `cargo test -p scarlet --test vm_io`: 2 passed (the `classify_overflow` unit tests), 25 ignored
- `cargo test -p scarlet --test vm_tls`: 2 passed, 21 ignored


alii:
Squashed into `rewrite` as d5c497f (`park two network tests`), one commit per PR. It lands on master with #42.

PR #78, opened 2026-09-20
the program carries no compiler ids

alii:
Stacked on #77. **The compiled `Program` no longer holds ids that only the compiler can read.** This gets the Core IR types ready to move into `scarlet_ir` next as a pure move. The VM should never be handed a number into a table it can't see.

## What changes

- **A constructor's `VariantRef` loses `type_name: StrId`.** It was written and never read. The comment said the display name is "looked up at emit", and the emitter is gone. It only made two `VariantRef`s for the same constructor compare unequal if they had been interned differently.
- **`CoreFn` loses `name: StrId`.** A function's readable name is already on `LoweredFn.name`, and the compiler has the id in its own typed program (`TypedFn.name`). So it carries the id beside the body while lowering (`LoweredBody { name, .. }`), and nothing puts it in the IR.
  - `lower`'s debug check for a misplaced global now says `fn#3` instead of `fn s1234`.
- **`LoweredFn.module` is the module key as text** (`"main"`, `"scarlet/array"`), not a `ModuleKey`. `ModuleKey` lives in `scarlet_syntax`, next to the file-path logic that builds it, and `scarlet_ir` depends on nothing. The Program only uses the module to name functions and to pick out one module's functions (`dis`, the snapshot tests). Those compare against `ModuleKey::main().as_str()`, so the text is always a real canonical key.

## Core IR snapshots

All 15 `.core` snapshots change, **in headers only**. I checked: 0 changed lines that aren't a `fn …(` line.

```
-fn s0(%0:t0) -> :t1
+fn main.sq(%0:t0) -> :t1
```

The harness prints each whole function now, under `module.name`, instead of the id. The renumbering of `s` ids goes, since no `s` ids are printed any more. An unnamed `CoreFn` prints as `fn(…)`.

## Checks

- `cargo test --workspace`: 892 passed, 0 failed, 323 ignored. The same tests as #77; the two network tests it parked account for the difference.
- `cargo hawk check -D warnings`: 0 findings. `LoweredFn::core` goes to `pub(crate)`, because the snapshot harness now prints the whole `LoweredFn`. It goes back to `pub` when it moves crates.


alii:
Squashed into `rewrite` as df82f5e (`programs carry no compiler ids`), one commit per PR. It lands on master with #42.

PR #79, opened 2026-09-20
core ir moves to ir

alii:
Stacked on #78. **The Core IR moves into `scarlet_ir`.** With this, the whole compiled `Program` (every type a backend reads) lives in the contract crate, and the compiler just produces it. This is the last "move the types" PR before the VM itself.

## What moves

- `scarlet_core/src/core_ir/mod.rs` → `scarlet_ir/src/core_ir/mod.rs`: `Program`, `LoweredFn`, `CoreFn`, `CoreExpr`, `Atom`, `CorePat`, `CoreBind`, `Callee`, `Load`, `VariantRef`, `ReuseShape`, `Const`, the index types (`LocalId`, `ConstId`, `FuncIdx`, `JoinId`), and their printing.
- `prim.rs` (`PrimOp`), unchanged (`git mv`).
- `GlobalSlot`, `FrameSlot` and `CaptureIdx`, from `typed_ir`: a `Load` names them, so they're part of the contract.

**What stays in the compiler:** the `lower` and `perceus` passes, `CoreProgram` (lower's own output), and the test builders. The compiler's `core_ir` module re-exports the moved types, so every `crate::core_ir::…` path still works.

**Reviewing the move:** git shows the new file as added, because `scarlet_core/src/core_ir/mod.rs` still exists, now as the compiler's small remainder. To see only what changed in the moved code:

```
git diff HEAD~1:crates/scarlet_core/src/core_ir/mod.rs HEAD:crates/scarlet_ir/src/core_ir/mod.rs
```

That's 182 changed lines, all of them one of these:

- `pub(crate)` → `pub`, and some private items → `pub`. `lower` and `perceus` used to reach them as child modules; now they're in another crate. That's `CoreBind`'s `id`/`ty`/`global`, `ReuseShape::fields`, `CoreBind::new`, `Atom::prim`, and so on. Hawk then narrowed back everything that turned out not to need it: `Atom::operands` is private again.
- `Indented` becomes `pub`, because the compiler's `CoreProgram` prints with it.
- The IR's own tests get two tiny local builders (`bind`, `vref`) instead of the compiler's test module.
- Doc comments that described the old emitter are rewritten as the contract: `Op::Reuse`, `MakeEnumPayload`, `PushLocal; MakeClosure`, `CallKnown`, `emit_toplevel`, and the module doc's "type erasure at Core→bytecode".
- The code moved out: `CoreProgram` and the test builders.

## One design point: `LoweredFn::pool`

`pool` is what every type in a function's body indexes, and nothing reads it until the VM does. #73 kept it with `#[expect(dead_code)]`. Moving the file made it `pub`, the expectation stopped holding, and the build said so, which is the point of `expect`. It's now **private**, and the compiler builds a `LoweredFn` with `LoweredFn::new(module, name, core, pool)`. So nothing outside `scarlet_ir` can read the pool until a backend does, and the `expect` note is true again.

## Checks

- `cargo test --workspace`: 892 passed, 0 failed, 323 ignored (unchanged from #78)

## Next

The VM crate: `scarlet_vm`, depending only on `scarlet_ir`. Its first PR runs `pub fn main() { println(1 + 2) }`.


alii:
Squashed into `rewrite` as 3bda134 (`core ir moves to ir`), one commit per PR. It lands on master with #42.

PR #80, opened 2026-09-20
the vm runs one plus two

alii:
Stacked on #79. **The new VM runs its first programs.** `scarlet run` on

```
pub fn main() {
	println(1 + 2)
}
```

prints `3`. That's step 2 of `docs/vm-design.md`. The VM lives in a new crate, `scarlet_vm`, which depends only on `scarlet_ir`.

## What it runs

- Int arithmetic and comparisons, with `docs/semantics.md`'s rules: `-7 / 2 == -3`, `x / 0 == 0`, `-7 % 2 == -1`, `x % 0 == x`.
- Calls to known functions, with arguments and results. A call in tail position reuses its frame.
- Module toplevels in order, globals, and `main`.
- `println` of an Int, a Bool or `Nil`.

Anything else **stops the run and says what it needed**, for example `cannot run: the new VM does not run String yet`. It never crashes and never guesses. A function that uses something unbuilt still loads, and only a *call* to it stops the run, so a program runs as far as the VM has come.

## How it's built

Three small files:

- **`value.rs`: the value word.** Every value is one 64-bit word. A float is stored as itself. Every other value uses bit patterns only a negative NaN could have, and Scarlet never stores a NaN, so nothing collides. The table of tags is at the top of the file. Small ints are 48-bit (±2^47). `Value` is `Copy`: when heap values come, their counts are kept by the VM against the process's own heap, not by Rust's `Clone`/`Drop`. There's no `PartialEq`, because comparing bits isn't Scarlet's `==` (`0.0 == -0.0`).
- **`code.rs`: instructions, and the loader.** Each function's Core IR becomes flat register instructions once, at load. A register *is* a Core IR local: `let %2 = IntAdd(%0, %1)` becomes `Int { dst: 2, op: Add, a: 0, b: 1 }`.
- **`exec.rs`: the loop.** Frames live on a stack the VM owns, not Rust's, and all frames' registers share one vector.

**An Int past 48 bits stops the run** (`Int beyond 48 bits (big ints)`), because big ints aren't built yet. Wrapping would print a wrong number.

**A closed output**, like `| head`, stops the run quietly with exit 0. That's `docs/vm-design.md`'s rule for the old VM's closed-pipe crash.

## Tests

- `scarlet_vm`: 4 unit tests on the value word, including that no storable float decodes as a tagged word. There are also 8 end-to-end tests that compile Scarlet source and check exact output: calls, the division rules, comparisons, `Nil`, only a *call* to an unbuilt function stopping, the 48-bit stop, and a closed output. The compiler is a dev-dependency only, so the VM itself still sees nothing but `scarlet_ir`.
- **13 parked tests pass now, and are un-parked:** 5 test functions and 8 `run_case!` cases. To un-park cases one at a time, the `needs the VM` ignore moved from the `run_case!` macro onto each of its 69 remaining cases.
- The two network tests from #77 also "passed", but only by skipping themselves again, so they stay parked.

## What the parked tests say comes next

When I ran all the parked tests, this is where they stopped:

| stops at | tests |
|---|---|
| Int beyond 48 bits | 119 |
| constructors | 46 |
| String | 30 |
| arrays (`MakeArray`) | 16 |
| `if` | 12 |

The 119 are mostly *not* the tests' own numbers. They're 64-bit constants in stdlib module toplevels, like `int.max_value`, which run before the test's code does. So I've moved **big ints up** in the design doc's order of work. It's now: control flow, the per-process heap, big ints, then constructors.

## Also

- The repo's instructions file: its crate-layout note describes `scarlet_vm`, and it now says 44 test executables (it was 42).
- `TiVec::get` comes back. The VM looks up functions with it rather than indexing, which could panic.

## Checks

- `cargo test --workspace`: **917 passed**, 0 failed, **310 ignored** (down from 323)
- `cargo dylint --all` with `DYLINT_RUSTFLAGS=-D warnings`: passes. It first caught a catch-all arm over the VM's own `View` enum, now written out variant by variant.


alii:
Squashed into `rewrite` as 3b79fdb (`vm runs one plus two`), one commit per PR. It lands on master with #42.

PR #81, opened 2026-09-20
the vm branches

alii:
Stacked on #80. **The VM runs `if`**, in both places it appears: as a function's result, and as a value inside an expression (`x = if a > b { a } else { b }`). It also runs calls to `self`. So loops written as recursion work, and so does `examples/bench.scrl`, the first golden example to pass on the new VM:

```
fn fib(n) {
	if n < 2 { n } else { fib(n - 1) + fib(n - 2) }
}
fn count(n, acc) {
	if n == 0 { acc } else { count(n - 1, acc + 1) }
}
pub fn main() {
	println(fib(26))
	println(count(200000, 0))
}
```

## How

Two new instructions: `Jump { to }` and `JumpIfFalse { cond, to }`.

The loader now knows where each expression's value goes (`Dest`):

- **`Return`:** a `Tail` returns, or tail-calls.
- **`Join { reg, after }`:** inside a `LetJoin`, a `Tail` writes the branch's value into the join's register and jumps past the branch. A call there is an ordinary call, not a tail call.

`If` is a `JumpIfFalse` to the else branch, then both branches with the same `Dest`. Jumps to code not placed yet are patched when their label is placed (`Label`, `place`, `jump`).

`call self(…)` resolves to the function's own index at load. A toplevel has no index, and never calls itself.

## Tests

- New `scarlet_vm` tests:
  - `if` in both positions
  - a million-step tail-recursive loop
  - **a 200,000-deep recursion that is not a tail call.** It passes because frames live on the VM's own stack, not Rust's: deep recursion uses memory and can't overflow anything, which is `docs/vm-design.md`'s "never recurse in Rust" rule working.
- Un-parked: the `bench` golden and `bench_runs`. To make that possible, the golden `suite!` macro now takes an attribute per entry, like `run_case!` in #80, so each parked program carries its own `#[ignore = "needs the VM"]`.

## Checks

- `cargo test --workspace`: **922 passed**, 0 failed, **308 ignored** (down from 310)
- `cargo hawk check -D warnings`: 0 findings. `TiVec::next_idx` is `pub` again, because the loader asks each function its own index.

## Where the parked tests stop now

Big ints (119) still lead, then constructors (54, up from 46, because tests that stopped at `if` now get further), then strings (30).


alii:
Squashed into `rewrite` as d03fcbd (`the vm branches`), one commit per PR. It lands on master with #42.

PR #82, opened 2026-09-21
write down heap growth

alii:
Stacked on #81. **Adds how a process heap grows to `docs/vm-design.md`, marked Proposed.** No code changes.

## The rule

- **No heap until it's needed.** A process gets no chunk until its first heap value, so one that only does Int maths costs nothing here.
- **Small values fill chunks.** They start around 2 KB and grow about 1.6 times per chunk. That keeps the chunk count small: a 1 MB list of 24-byte cells fits in 13 chunks, not about 43,000.
- **A value bigger than the current chunk size gets a chunk of exactly its size.**
- **An empty chunk goes back to the system.**

## Where the numbers come from

It's BEAM's design, checked against Erlang/OTP `master` at `bf52b4d716` (2026-09-21), with file names in the doc:

- a new heap is 233 words (`H_DEFAULT_SIZE`), and a whole new process is 327 words, about 2.6 KB
- heap sizes are Fibonacci plus 1 word, about 1.6 times per step, up to 833,026 words, then +20% per step
- when a heap is short, BEAM allocates a heap fragment of *exactly* the size needed (`erts_heap_alloc`), and messages that can't go straight into the receiver use one too

BEAM rounds up to its size table only when its copying GC puts everything back into one block. We never move values, so we take the exact-size part and skip the copying part.

## New open question

Fragmentation: freed cells leave gaps we can't squeeze out. BEAM hit this too: shrinking hibernated heaps in place "caused serious fragmentation problems", so `hibernate` now copies into an exact-size heap. The doc says to measure the gaps once processes exist.

## Also

The memory-limit bullet now says that a process which passes its limit dies, and **its links hear about it** like any other death. That's how BEAM's `max_heap_size` works too.


alii:
Squashed into `rewrite` as 3f2554c (`write down heap growth`), one commit per PR. It lands on master with #42.

PR #83, opened 2026-09-21
a heap and hello world

alii:
Stacked on #82. **The VM gets a heap, reference counting, and its first heap value: strings.** `examples/hello.scrl` runs and matches its golden file:

```
hello, world
hello from Scarlet
2 + 2 = 4
```

This is step 4 of `docs/vm-design.md`, for one process. Per-process limits come when there's more than one process.

## The heap (`heap.rs`)

It follows #82's rules:

- **No chunk until the first cell.** A program that never makes a heap value never allocates a chunk.
- **Small cells fill chunks** of 256 words (2 KB), then 409, 654, 1046, and so on: 1.6 times each.
- **A cell bigger than the current chunk size gets a chunk of exactly its size**, and that chunk goes back to the system when the cell is freed.
- **A freed small cell goes on a free list**, by size, for the next cell of that size.

Every cell starts with one header word: its count, its kind and its size. A cell is named by chunk and word, not by address, so the heap needs **no `unsafe`** and moves with its process as plain data. A cell fits in the value word's 48 bits: 20 of chunk, 28 of word. If a heap outgrows that, the run stops with the new `Stop::HeapFull`, which is a limit of the machine, not a bug.

Not yet: a small-cell chunk whose cells are all free stays allocated. Only big cells' own chunks go back so far.

## Reference counting (`exec.rs`)

One rule, stated at the top of the file: every register, global and argument holds one reference. So:

- copying a value into a register adds a reference
- overwriting a register, or leaving a frame, gives one up
- Perceus's `Drop`, now a real instruction, gives one up early

That keeps every count right, whether or not Perceus inserted a drop. This matters because Perceus never drops `String`.

**How I know counts don't leak:** three tests run a program, release what a finished run still holds (its result and its globals), and check that the heap has **0 cells left**:

- strings made, joined and returned through calls
- a 10,000-step loop that makes a new string each step and drops the last
- a string held in a global

## Strings

- A string constant is made when its instruction runs. It doesn't get a shared area until constants do.
- `+` and interpolation (`StringConcat`, `StringConcatMany`) join strings.
- `${x}` (`ToString`) works for Int, Bool, `Nil` and String.
- `println` prints strings.
- Not yet: `True`/`Nil` written in source (they lower as constructors), and the stdlib's string functions.

## Tests

- `scarlet_vm`: 7 new heap unit tests (no chunk before the first cell, reading back at every length, reuse, sharing, the growth sizes, a big cell's own chunk, 48-bit round trip), 3 leak tests, and a strings end-to-end test.
- **Un-parked 8:** the `hello` golden, five multi-file module tests (programs split across files now run), and two type tests.

## Also

`docs/vm-design.md`'s order of work marks steps 3 (`if`) and 4 done. It notes that `match` comes with constructors, since most matches are over them, and that 300 tests are left parked.

## Checks

- `cargo test --workspace`: **941 passed**, 0 failed, **300 ignored** (down from 308)


alii:
Squashed into `rewrite` as 6fb7332 (`a heap and hello world`), one commit per PR. It lands on master with #42.

PR #84, opened 2026-09-21
ints never overflow

alii:
Stacked on #83. **Ints are exact at any size.** Past 48 bits an Int moves to the heap, and past 64 bits it keeps going. That's step 5 of `docs/vm-design.md`, and it makes the "Int is arbitrary precision" decision in `docs/semantics.md` real in the VM.

```
println(9223372036854775807 + 1)   // 9223372036854775808
println(fact(25))                  // 15511210043330985984000000
```

## How

- **One form per number.** An Int that fits in 48 bits lives in the value word; any other is a heap cell. Every result is put back in that form, so a result that fits is always small (`bigint::value`). A small Int and a big one are never the same number, so `==` can't be fooled.
- **Fast path first.** Two small operands use a checked `i64` operation. Only when the answer needs more room does the arithmetic move to `num-bigint`, which the design doc proposed, behind our own small `bigint.rs`.
- **The rules hold for big ints too:** `/` truncates toward zero, `x / 0 == 0`, `%` takes the dividend's sign, and `x % 0 == x`. `num-bigint` **panics** on a zero divisor, so the zero cases are handled before it's ever called.
- **The heap cell (`Kind::BigInt`):** after the header, one word holds the sign (bit 63) and the digit count, then the 64-bit digits, least significant first.
- **A constant** past a small Int's range becomes a `BigInt` instruction, made when it runs, like string constants in #83.

## Not yet

An Int *literal* past 64 bits, like `1000000000000000000000000000000`, is still a **compile error**. The compiler keeps Int constants as `i64` (`Const::Int(i64)`). That's the compiler's to fix, in its own PR. Big numbers can already be computed, just not written down.

## What the parked tests say

Only 1 more passes (`stdlib_net_socket_type`). The 119 that stopped at big ints all got further, and all stop at **constructors** now. Stdlib module toplevels build constructor values (`Some`, `Ok`, records) before any test code runs. So constructors are clearly next, and the design doc says so.

## Tests

- `scarlet_vm` end-to-end:
  - exact past 48 and 64 bits (factorial of 25, `int.max_value + 1`, `int.min_value`)
  - a big result that fits comes back small (`big / 4 == 9223372036854775807`, `big - big == 0`)
  - the division and zero rules on big ints
- A heap test that a big int reads back with its sign.
- A leak test: 200 ever-bigger numbers in a loop, then printed and divided, leave **0 cells** behind.

## Checks

- `cargo test --workspace`: **946 passed**, 0 failed, **299 ignored**


alii:
Squashed into `rewrite` as 88df937 (`ints never overflow`), one commit per PR. It lands on master with #42.

PR #85, opened 2026-09-21
bool and nil are immediates

alii:
Stacked on #84. **`True`, `False` and `Nil` are plain values in the Core IR, never constructors.** A `match` on a Bool is an `if`. This comes before constructors in the VM, so the VM never has to know which type is the prelude's `Bool`.

## Why

A Bool could look two ways in the Core IR:

- `x < y`, `a && b` → `Atom::Bool` (a plain value)
- `True`, `False` → `ctor 261.0()`, `ctor 261.1()` (the prelude's constructor, type id 261)
- the same for `Nil`: an empty block gives `nil`, but `Nil` gives `ctor 262.0()`

So a backend would need to know that type 261 is `Bool`, and treat a `True` built one way the same as a `True` built the other way. Now there is only one way.

## How

A new compiler pass, `core_ir/immediates.rs`, runs right after `lower` on every function and on the toplevel:

- `ctor True()` / `ctor False()` / `ctor Nil()` → `true` / `false` / `nil`
- `match b { True -> x, False -> y }` → `if b { x } else { y }`. Wildcards count, and the arm that runs first wins, just like `match`.
- `match n { Nil -> x }` → just `x`, since there is only one Nil.
- An arm that names the value (`x -> x`) counts as a catch-all, and starts with `let x = b`.
- Anything else is left alone: a match over another type, or a match of only `_` and names.

It runs **between `lower` and Perceus**. So Perceus never sees a Bool as a constructor, and never marks a Bool's drop as a cell to reuse (`drop %0 [ctor:0]`). That mark is how I found the `x -> x` case: at first this pass skipped such matches, and Perceus marked them.

It takes the prelude's Bool and Nil from `self.prelude`, the same place the rest of the compiler gets them.

## What the parked tests say

**6 more pass**:

- `stdlib_bool`
- `and_or_short_circuit_skips_rhs`, `and_or_evaluate_rhs_when_lhs_undecided`
- `mutual_recursion_functions`, `mutual_tail_recursion_between_known_fns`
- `mutually_recursive_scc_jumps_over_every_parked_body`

Two network tests (`connect_within_…` and `connect_addr_within_times_out_against_a_full_accept_queue`) *also* pass on a Mac, but only because they skip themselves when they can't fill the accept queue. They'd really run on Linux CI, so they stay parked.

## Tests

- Unit tests for the pass:
  - the three constructors become values
  - a Bool match becomes an `if` with the `True` arm as `then`, and with a wildcard too
  - a Nil match becomes its arm
  - an `x -> ...` arm binds the value
  - other matches, and a match of only `_`, stay
- A new Core IR snapshot, `bool_match_is_if`, with an `x -> x` arm.
- Two snapshots change one line each: `ctor 261.0()` → `true`.

## Checks

- `cargo test --workspace`: **957 passed**, 0 failed, **294 ignored**. One more is ignored than below because #77 now parks a Linux-only network test that CI had been failing on since #77.


alii:
Squashed into `rewrite` as 569e432 (`bool and nil are immediates`), one commit per PR. It lands on master with #42.

PR #86, opened 2026-09-21
programs name their types

alii:
Stacked on #85. **A `Program` now carries the names of its types.** That's what the VM needs to print a constructor: `Some(1)`, `Err(Nil)`, `Point{ x: 1, y: 2 }`. The VM itself comes in the next PR. This one only adds the names and uses them in `scarlet dis`.

## Why a table

In the Core IR a constructor is only numbers: `ctor 263.0(%1)` is type 263, variant 0. That's the right thing to *run*, but a person reading output needs `Some`. The old VM kept names in its constructor templates. The new VM only gets a `Program`, so the names have to travel in it.

## What's in it

`Program::types: BTreeMap<TypeId, TypeNames>` (in `scarlet_ir`):

- `TypeNames { name, variants }`: the type's name as declared
- `VariantNames { name, fields }`: each variant's name and its field labels, in the order the constructor holds them

It holds **every type the program builds or matches a constructor of**, and only those. The compiler fills it in `finish_program`: it walks every body (functions, module inits, toplevel) with the new `CoreExpr::for_each_variant`, then reads the names from the type environment, by `TypeId`. A constructor whose type has no variants there is an internal compiler error, like a function that never lowered.

It's keyed by `TypeId`, which is the type's identity, not by a name, since two modules can each have a `Method`.

## `scarlet dis`

A listing now ends with the types its functions use:

```
; types
; 263 Option: .0 Some(value), .1 None
; 768 Point: .0 Point(x, y)
```

So `ctor 263.0(%1)` reads as `Some(%1)`.

## Also

The design doc's table of `Program` fields gets a `types` row. It also pointed at the old home of `Program` (`scarlet_core`); that's `scarlet_ir` since #79.

## Tests

- `every_constructor_has_names`: a program importing `http`, `json`, `map` and `string` has names for every constructor in every body, down to the variant (over 100 constructors), including its own `Shape`.
- `dis`: a listing ends with its own types, and only those.

## Checks

- `cargo test --workspace`: **959 passed**, 0 failed, **294 ignored**


alii:
Squashed into `rewrite` as beefafb (`programs name their types`), one commit per PR. It lands on master with #42.

PR #87, opened 2026-09-21
tail calls own their arguments

alii:
Stacked on #86. **Perceus stops dropping a tail call's arguments before the call.** Found while building constructors in the VM (the next PR): a list built by a tail-recursive loop came out as `Nil`.

## The bug

```
fn build(n Int, acc L) L {
	if n == 0 { acc } else { build(n - 1, Cons('${n}', acc)) }
}
```

Perceus turned the loop's last line into:

```
let %7 = ctor Cons(%6, %1)
drop %1
drop %7              <- gives up the new list...
ret call self(%5, %7) <- ...before the call reads it
```

Read in order, that frees `%7` and then passes it on. The old VM's self tail call kept its frame and needed that drop, so its emitter (`split_self_tail_drops`) moved it past the argument reads. The new VM runs the Core IR in the order it's written, so the call got `Nil`.

## The fix

`release_self_tail_args` is gone. A tail call in the new VM reads its arguments, then gives up every reference its frame still holds, because the callee takes the frame's place. So the drop isn't needed at all. That's already how every tail call to *another* function worked. Now a call to itself works the same way.

Other locals still drop before the tail call as usual (`drop %1` above stays).

It only showed up now because Perceus only drops constructor, tuple and function values (`is_heap`). Strings and Ints never get a `Drop`, so the string loops in earlier PRs didn't hit it.

The design doc's memory rules get this as one more line.

## Tests

- `a_tail_call_argument_is_not_dropped_before_the_call`, for a call to itself and to another function. I checked it **fails on the old Perceus**, which puts `drop %1` right before the call.
- No Core IR snapshot changes: none had a heap argument in a self tail call.

## Checks

- `cargo test --workspace`: **960 passed**, 0 failed, **294 ignored**


alii:
Squashed into `rewrite` as 327abe3 (`tail calls own their arguments`), one commit per PR. It lands on master with #42.

PR #88, opened 2026-09-21
the vm builds constructors

alii:
Stacked on #87. **The VM builds, frees and prints constructors.** This is the first half of step 6 in `docs/vm-design.md`. `match` is the other half, next.

```
println(Some(1))                    // Some(1)
println(Err(Nil))                   // Err(Nil)
println(Point(x: 1, y: -2))         // Point{ x: 1, y: -2 }
println(Seg(a: Point(..), b: ..))   // Seg {
                                    //   a: Point{ x: 1, y: 2 },
                                    //   b: Point{ x: 3, y: 4 }
                                    // }
```

## How a constructor is stored

- **With fields** (`Some(1)`): a heap cell (`Kind::Ctor`). After the header, one word holds the type (bits 0 to 31) and variant (32 to 47), then one word per field. Each field holds one reference, like a register does.
- **Without fields** (`None`, `Dot`): no cell at all. It's a value word of its own (tag `0xFFFD`), holding its type and variant, like `True` is. So it's never counted or freed.

Perceus's `reuse` hint is ignored for now. A fresh cell is always correct, and reuse comes in its own PR.

## Freeing

When a constructor's count reaches zero, each of its fields loses a reference, and so on down. That uses a **work list, not recursion**, so freeing a million-long list takes no more Rust stack than freeing one cell. The list is kept on the heap between calls, so a release doesn't allocate.

## Printing

`show.rs` uses the old VM's layout rules (`inspect.rs` on master), so programs print what they always printed:

- `Name(a, b)` for a variant, and `Name{ x: 1, y: 2 }` for a record, meaning a constructor named after its own type.
- If any field isn't small, one field per line, indented. Small means an Int, Bool, Nil, a constructor with no fields, or a string under 20 bytes.
- Names come from `Program::types` (#86).

It's also a work list, so a deeply nested value can't overflow the stack.

## What the parked tests say

Only **1 more passes**: `aliased_type_import_unifies_with_canonical`, which prints a constructor from another module. The rest now stop further on:

- 62 at `match`
- 16 at its jumps (`LetCont`/`Goto`)
- 14 at a record field (`FieldUnchecked`)

So `match` is next.

## Tests

- `scarlet_vm` end to end: variants, records, `Some(None)`, `Err(Nil)`, `${Some(True)}`, the one-per-line layout, and a 100,000-long list built by a loop.
- Leak test: lists of strings, a record in a global, and `Some(origin)` leave **0 cells**.
- Heap: a million-long chain freed by one release, with a shared string inside that survives.
- Value: a constructor without fields keeps its type and variant, including negative type ids and the largest variant number.

## Checks

- `cargo test --workspace`: **969 passed**, 0 failed, **293 ignored**


alii:
Squashed into `rewrite` as 7f306eb (`the vm builds constructors`), one commit per PR. It lands on master with #42.

PR #90, opened 2026-09-21
the vm matches

alii:
Stacked on #88. **The VM runs `match`.** It covers constructors, Int and string literals, nested patterns, and record fields (`p.x`). That finishes the constructors half of step 6. **45 parked tests pass.**

```
fn area(s Shape) Int {
	match s {
		Circle(r) -> 3 * r * r
		Rect(w, h) -> w * h
		Dot -> 0
	}
}
```

## How a match loads

Each arm becomes a test, in order. If the value doesn't fit, it jumps to the next arm's test. If it does, it falls into the arm:

```
JumpUnlessVariant %0 Circle -> next
Field %1 = %0.0          <- what the pattern binds
...arm...
next:
JumpUnlessVariant %0 Rect -> next2
...
```

- `Circle(r)`: `JumpUnlessVariant`, then one `Field` per field. Each field read takes its own reference, like any register.
- `0`, `'hi'`: `JumpUnlessInt` / `JumpUnlessStr`. The Int test also compares a big int exactly.
- `x`: a `Move`. `_`: nothing.
- A nested pattern is already a match inside a match (`lower` does that), and a case that fails part-way uses `LetCont`/`Goto`. The continuation goes after its body, and a `Goto` is a jump to it.
- `p.x` (`FieldUnchecked`) is a `Field` too.

Tests are in arm order, so the first arm that fits wins, like in the source. A jump table could come later, with the JIT.

## A new stop: `BadProgram`

After the last arm the loader puts an `Instr::Bad`. The compiler promises every match has an arm for every value, so only a compiler bug gets there. It stops the run with `Stop::BadProgram` and says what happened, instead of running on into the wrong code. The driver prints `internal error: ... This is a bug in the compiler, not in the program`.

A few things that were wrongly reported as "not built yet" use it now too:

- a `Goto` to no case
- a field read off a value without that field
- a function running past its last instruction

## What the parked tests say

45 pass: constructor patterns and destructuring, qualified and aliased constructors across modules, guards, or-patterns, field access, exhaustive matches, the `pipe` example program, and `bench_typed.scrl`'s output. No test hit `BadProgram`.

248 are left, and what they stop at is spread out now:

- arrays: 32
- binaries: 31
- calling a function value: 21
- processes: 20
- captures, tuples and floats: about 10 each

## Tests

- `scarlet_vm` end to end:
  - variants with fields
  - nested `Option(Option(Int))` falling through part-way
  - Int, or-pattern and string arms, including `9223372036854775807`
  - `p.x`
  - a list summed by recursion, and one 100,000 long walked by a loop
- Leak test: building, reversing and reading lists of strings leaves **0 cells**.
- Heap: `string_is` against the text, one byte more, and one byte changed, at lengths around the 8-byte word.

## Checks

- `cargo test --workspace`: **1021 passed**, 0 failed, **248 ignored**


alii:
Squashed into `rewrite` as 20c2394 (`the vm matches`), one commit per PR. It lands on master with #42.

PR #91, opened 2026-09-21
the vm runs closures

alii:
Stacked on #90. **The VM runs closures and calls function values.** That's step 7 of `docs/vm-design.md`. **21 parked tests pass**, including `stdlib_option` and `stdlib_result`.

```
fn adder(n Int) fn(Int) Int { fn(x) { x + n } }

pub fn main() {
	println(adder(2)(4))                                  // 6
	k = 7
	go = fn(n) { if n == 0 { k } else { go(n - 1) } }
	println(go(100000))                                   // 7, in constant space
	println(adder)                                        // <fn#adder>
}
```

## How a closure is stored

- **With captures:** a heap cell (`Kind::Closure`). After the header, one word holds its `FuncIdx`, then one word per capture. Each capture holds one reference, and freeing the closure frees them, using the same work list as constructors.
- **Without captures:** no cell, as before. It's the function's own value word.

## Calling

A frame now remembers **the closure it runs as** (`Frame::env`), and holds one reference to it. `capture0` reads from it.

A call names its callee one of three ways, and each has its own instruction, with a tail form that replaces the frame:

- `Call` / `TailCall`: a function known at load time. It runs with no closure.
- `CallSelf` / `TailCallSelf`: the function running now, **as the same closure**. That matters: a lambda that captures and calls itself (`go` above) is `call self` in Core IR, so the call has to keep its captures.
- `CallValue` / `TailCallValue`: a function value in a register, with or without captures. A tail call takes its own reference to the closure *before* the frame's registers are released, since one of them may be the only other holder.

A call to something that isn't a function, or with the wrong number of arguments, is `Stop::BadProgram` (#90). The compiler rules both out.

`Load::SelfClosure` (a lambda reading itself as a value) gives the frame's closure, or the plain function value when it runs without one. `Load::Slot`, the selective-import escape hatch, is still "not built yet": I couldn't find a program that makes one (a selective import turns into a global).

## Printing

A function prints as `<fn#name>`, whatever it captured, like the old VM. So `<fn#serve>` in the REPL test, and `<fn#__anon__>` for a lambda.

## Also in the design doc

Two decisions from today's discussion, written into the plan:

- **Arrays** stay the RRB tree the old VM had. Scarlet code uses arrays like lists: walking with `[h, ..t]`, appending in `array.map`, prepending in `reverse`. Only the tree is fast at all three. `[h, ..t]` should give a view, not a copy.
- **`wire`**: the descriptor builder (`typed_ir/wire.rs` on master) was removed with the old VM and has to come back. Today's Core IR says only `WireDecode(%0)`. A closure on the wire will name **its code, not its run**: its module, its function and a fingerprint of the module's code, as BEAM's `NEW_FUN_EXT` does with the module's MD5. So it can be decoded after a restart.

## What the parked tests say

21 pass: captures of every kind, recursive and shadowed lambdas, calls through values including a deep tail call (`indirect_tail_call_through_value_reuses_frame`), `stdlib_option`, `stdlib_result`, `stdlib_resource`, and the `backpassing` example program. None hit `BadProgram`.

227 are left:

- arrays: 33
- binaries: 31
- processes: about 24
- `wire`: 17
- floats: 13
- tuples: 8

## Tests

- `scarlet_vm` end to end:
  - captures, including a string
  - closures returned from a call and called later (`adder(2)(4)`, `twice(adder(10), 0)`)
  - a capturing lambda calling itself, 100,000 deep in tail position and non-tail too
  - a lambda returning a lambda
  - named functions as values
  - printing a function, and a function inside `Some`
- Leak test: closures made in a loop of 1,000, holding records of strings, returned, called and dropped, leave **0 cells**.
- Heap: a closure holds its captures, and freeing it frees them.

## Checks

- `cargo test --workspace`: **1048 passed**, 0 failed, **227 ignored**


alii:
Squashed into `rewrite` as ddd24e5 (`the vm runs closures`), one commit per PR. It lands on master with #42.

PR #92, opened 2026-09-21
the vm builds tuples

alii:
Stacked on #91. **The VM builds, reads and prints tuples.** A small one. **5 parked tests pass**.

```
fn swap(t (Int, String)) (String, Int) { (t.1, t.0) }

println(swap((1, 'one')))   // (one, 1)
println((1, (2, 3)))        // (
                            //   1,
                            //   (2, 3)
                            // )
```

## How

- **Stored as** a heap cell (`Kind::Tuple`): the elements' value words, right after the header. There's no tag word, since the size already says how many elements. Each element holds one reference, and freeing the tuple frees them with the same work list as constructors and closures.
- **`MakeTuple`** becomes a `Tuple` instruction, and **`t.0`** (`TupleField`) an `Element` instruction. A tuple pattern already lowers to element reads under a `_` arm, so matching needed nothing new.
- **Printing** uses the old VM's rules:
  - `(a, b)` on one line when every element is small and the line fits in 80 columns
  - otherwise one element per line
  - `()` when empty
  - a tuple is never "small" itself, so a constructor holding one takes a line per field, as before

The one-line attempt writes the small elements directly and backs out if the line gets too wide. A small value holds no other value, so that's bounded work and no recursion.

## What the parked tests say

5 pass: or-patterns inside tuples, a tuple index followed by a field read, and an aliased constructor nested in a tuple. 222 are left.

## Tests

- `scarlet_vm` end to end:
  - swapping by `t.1, t.0`
  - a nested tuple pattern
  - the one-per-line layout
  - `Some((True, Nil))`
  - `${(1, 2)}`
  - the 80-column boundary: exactly 80 stays on one line, 81 splits
- Leak test: tuples of strings rebuilt 1,000 times in a loop, read, printed and wrapped leave **0 cells**.
- Heap: a tuple holds its elements, the empty tuple too.
- `only_calling_an_unbuilt_function_stops_the_run` used a tuple as its "not built yet" example. It uses an array now.

## Checks

- `cargo test --workspace`: **1057 passed**, 0 failed, **222 ignored**


alii:
Squashed into `rewrite` as 9bf7e12 (`the vm builds tuples`), one commit per PR. It lands on master with #42.

PR #93, opened 2026-09-21
the vm builds arrays

alii:
Stacked on #92. **The VM builds arrays**, as the RRB tree the old VM had (`seq.rs` on master). This is the choice we made: Scarlet code uses arrays like lists (walking with `[h, ..t]`, appending in `array.map`, prepending in `reverse`), and only the tree is fast at all three.

```
fn total(xs Array(Int), acc Int) Int {
	match xs {
		[] -> acc
		[h, ..t] -> total(t, acc + h)
	}
}

println(total(build(10000, []), 0))   // 50005000
println([0, ..[1, 2], 3])             // [0, 1, 2, 3]
println([10, 20, 30][3] or 99)        // 99
```

## The tree (`array.rs`)

A **root** holds the length, the tree's height, and three parts:

- a **head** buffer leaf, where a push to the front lands
- the **tree**: branches of up to 32 children, each keeping running totals of how many elements are under it
- a **tail** buffer leaf, where a push to the back lands

Only a full buffer goes into the tree, once every 32 pushes, so a push at either end is cheap. The running totals make finding an element, cutting and joining one node visit per level. Joins repack a level when it would get more than 2 nodes looser than it could be (the RRB paper's rule), so lookups stay shallow however many joins built the array.

**Nothing changes in place.** An operation builds new nodes along the one path it touches and shares every other node with the array it came from, and that array stays as it was. Changing a node in place when nothing else holds it waits for "moves on last use".

The heap stays plain storage: `Heap::make` and `Heap::data` hand out raw words, and `array.rs` owns the layout. The heap only knows which words of a root, leaf or branch hold references, so freeing still follows them with its work list.

## What runs

- `[1, 2, 3]` (`MakeArray`)
- length (`ArrayLen`)
- patterns `[a, b, ..rest]` (`ArrayElem`, `ArrayDrop`)
- `[x, ..xs]` and `[..xs, x]` (`ArrayPrepend`, `ArrayAppend`)
- `[..a, ..b]` (`ArrayConcat`)
- `xs[i] or d` (`ArrayIndexOr`). A negative index or a big-int index has no element, so it gives `d`.

**Printing** follows the old rules: `[1, 2, 3]` when it fits in 80 columns, six to a line when it's all small values that don't fit, one element per line otherwise, and `[]` when empty. `inspect_multiline_structures_e2e`, the test that pins the old nested layout, passes.

## Not yet, on purpose

- **`xs[i]`** builds an `Option`, so the VM needs to know which constructors are the prelude's `Some` and `None`. The old VM had a table for that (`AbiSlot`), removed with it. That's the next PR.
- **Ranges** (`0..10`, `MakeRange`): the PR after.
- **`xs[a..b]`**: the old VM *crashed* when the range went past the end, and `docs/semantics.md` lists that crash as a bug to fix. What it should do instead is a language decision, so slicing stays "not built yet" until that's made. `take` is written and tested, and marked unused outside tests until then.
- **A cheap `[h, ..t]`**: today it copies the rest of the head buffer (up to 31 elements) per step. The plan is a view.

## What the parked tests say

Only 3 pass (`array_spread_literal`, `bare_builtin_as_value_is_callable`, `inspect_multiline_structures_e2e`). The ~33 array tests got further and now stop at:

- ranges: 11
- `string.inspect`: 9
- `xs[i]`: 6
- `==` on arrays: 6
- the `array.length` built-in: 5

219 are left.

## Tests

- **Tree** (`array.rs`), every invariant checked: node heights, slot counts, running totals and length.
  - arrays of 0 to 40,000 read back
  - 3,000 pushes at both ends leave the first array unchanged
  - `take`/`skip` at every boundary of a joined array
  - joins of every size against every size, and 300 joins in a row staying at most 3 levels tall
  - every test ends with 0 cells left
- **End to end:** printing (one line, six to a line, one per line, nested), walking with `[h, ..t]` by recursion and by a loop, 10,000 appends in a loop, `[a]`/`[a, b]`/`[a, _, ..rest]` patterns, spreads, joins, and `xs[i] or d` in range, past the end, negative, and at `9223372036854775807`.
- **Leak test:** 2,000 strings appended, 2,000 more prepended onto that same array, the two joined, walked and indexed, with the old versions kept around: **0 cells** left.

## Checks

- `cargo test --workspace`: **1068 passed**, 0 failed, **219 ignored**


alii:
Squashed into `rewrite` as 5074f62 (`the vm builds arrays`), one commit per PR. It lands on master with #42.

PR #94, opened 2026-09-21
the vm builds options

alii:
Stacked on #93. **The `Program` tells the VM which constructors are the stdlib's `Some` and `None`, and `xs[i]` runs.** 6 parked tests pass.

```
xs = ['a', 'b', 'c']
println(xs[0])    // Some(a)
println(xs[3])    // None
println(xs[-1])   // None
```

## Why a table

`xs[i]` returns an `Option`, so the VM has to build `Some(x)` or `None` itself. But the VM knows nothing about the stdlib's types, and shouldn't: `Option` is just type 263 in one compile and something else in the next. The old VM solved this with a table (`AbiSlot`) that the compiler filled in. It was removed along with the old VM.

This brings it back, through `scarlet_ir`:

```rust
pub struct Abi {
    pub some: VariantRef,
    pub none: VariantRef,
}
```

`Program::abi` is filled in `finish_program` from the prelude bindings the compiler already has (`self.prelude.some()` / `none()`). The prelude is always loaded, so the table is always complete and has no "not bound" case to check. It will grow one field at a time as the VM builds more: `Ok`/`Err` come next, with slices.

The types of the table's constructors are added to `Program::types`, so a `Some` the VM builds prints as `Some` even in a program that never writes `Some` itself.

Also: `CtorRef` → `VariantRef` is a `From` now, replacing a hand-written closure in `immediates()`.

## The VM

- `ArrayIndex` → an `ArrayIndex` instruction. It builds `Some(element)` as a cell, or `None` as the value word it already is.
- It shares the index rules with `xs[i] or d`: past the end, negative, or a big int is no element.

## Decided, and written down

**`xs[a..b]` returns `Result(Array(a), Nil)`**, with `Err(Nil)` when the range is out of bounds. The old VM crashed there. That's the next PR, and it's in `docs/vm-design.md` now.

## What the parked tests say

6 pass: `array_index_yields_option`, `index_returns_option`, `index_negative_returns_none`, `index_or_covers_both_encodings_and_every_boundary`, `index_or_does_not_evaluate_an_impure_default`, and `constructor_is_first_class`. 213 are left.

## Tests

- `scarlet_vm` end to end: `xs[i]` at both ends, past the end, negative, at `9223372036854775807`, matched with `Some(s)`/`None`, and with `or` using a constant default and a computed one.
- `every_constructor_has_names` also checks that the table's constructors have names, and that `none` really is `Option`'s `None`.

## Checks

- `cargo test --workspace`: **1075 passed**, 0 failed, **213 ignored**


alii:
Squashed into `rewrite` as 0ee43eb (`the vm builds options`), one commit per PR. It lands on master with #42.

PR #95, opened 2026-09-21
slices return a result

alii:
Stacked on #94. **`xs[a..b]` returns `Result(Array(a), Nil)`.** It's `Err(Nil)` when the range isn't inside the array, instead of a crash. This was decided in review: a slice can fail, so the failure should be a value the caller sees.

```
xs = [1, 2, 3, 4, 5]
println(xs[1..3])    // Ok([2, 3])
println(xs[3..9])    // Err(Nil)
println(xs[3..1])    // Err(Nil)
println(xs[2..2])    // Ok([])
```

## The rule (now in `docs/semantics.md`, "Arrays")

- `Ok` of the elements from `a` up to but not including `b`.
- `Err(Nil)` when `a` is negative, `b` is past the length, or `a` comes after `b`.
- `xs[a..a]` is `Ok([])`.

It's a `Result`, not a clamped shorter array, because a range that misses the array is a failure to show, not hide. `binary.slice_bits` already works this way. The old VM **crashed** here, and `docs/semantics.md` listed that crash as a bug. It's off that list now: two crash paths are left, both about processes.

## Changes

- **Type checker:** a range subscript is `Result(Array(elem), Nil)` instead of `Array(elem)`.
- **VM:** an `ArraySlice` instruction: take the first `b`, then skip the first `a`, using the tree code from #93. `take` isn't marked unused any more.
- **`Program::abi`** gets `Ok` and `Err` (#94 added `Some`/`None`).
- **Every slice in the repo** is updated. There were only five:
  - `examples/supervision.scrl` and `examples/chat.scrl` used an out-of-range slice **on purpose, to crash a process** and show a supervisor restart it. Under the rule, a supervisor is for deaths from outside the code, so they now do `process.kill(process.self())`, which `process.kill`'s doc allows and which restarts a supervised worker.
  - `chat.scrl`'s `remember` used `history[1..n]` to drop the oldest frame. It's a `[_, ..rest]` match now.
  - `tests/programs/semantics.scrl` prints the new results and two failing slices. Its golden output is updated. It's parked, since it needs `string.inspect`.
- **`u23`** in `unsound.rs` used to check that a bad slice crashed with a message. It now checks for `Err(Nil)`, runs through `scarlet run`, and passes, so it's unparked.

By the old print rules an array isn't "small", so `Ok([2, 3])` prints over three lines when it stands alone, like `Some([1])` does.

## Tests

- `scarlet_vm` end to end:
  - in range, whole, empty, empty at the end
  - past the end, backwards, a negative start, and an end of `9223372036854775807`
  - a slice deep in a 5,000-element tree, matched with `Ok`/`Err`
  - `or` on a failed slice
- `u23_oob_and_reversed_slice_are_err`, through `scarlet run`.
- `scarlet check` passes on both examples and `semantics.scrl`.

## Checks

All seven CI steps, since `.scrl` files changed:

- `cargo test --workspace`: **1077 passed**, 0 failed, **212 ignored**
- grammar: `bun install`, then `bun run generate && git diff --exit-code src grammar.js`, then `bun run corpus`: all pass


alii:
Squashed into `rewrite` as 3e6b43f (`slices return a result`), one commit per PR. It lands on master with #42.

